PR #4, opened 2026-02-25
[draft] handle reorgs

dhvanipa:
## Reorg Handling

The synchronizer tracks canonical beacon roots per slot (`slot_root:<slot>`) and checks them while syncing.

Reorg is detected when any of these happen:

- a slot that previously had a root is now missing
- stored root for a slot differs from newly fetched root
- fetched header `parent_root` does not match stored root of the previous slot

When detected, it:

1. walks backward to find the last slot where stored root matches live root
2. rewinds DB state to just before divergence
3. replays forward from the divergence slot

Rollback removes:

- tx/nullifier entries seen after the retained slot
- stored slot-root entries after the retained slot
- sync progress after the retained slot

### DIFF
--- a/synchronizer/src/db.rs
+++ b/synchronizer/src/db.rs
@@ -150,6 +150,7 @@ impl Db {
     pub fn rollback_to_slot(&self, keep_slot: Option<u32>) -> Result<()> {
         let mut batch = WriteBatch::default();
 
+        // Remove any derived records written after the retained canonical slot.
         for entry in self.db.iterator(IteratorMode::Start) {
             let (key, value) = entry?;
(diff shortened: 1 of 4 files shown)


robknight:
I think this would benefit from more documentation, either comments in the code, a README, or some other documentation which explains the design and how it works.

dhvanipa:
> I think this would benefit from more documentation, either comments in the code, a README, or some other documentation which explains the design and how it works.

Added some in-line comments and also updated the PR description with the general design. Let me know if things are still unclear!

dhvanipa:
superseded by https://github.com/dobjlabs/zk-craft/pull/18

PR #5, opened 2026-02-25
Add TX model support to synchronizer

robknight:
This adds support for the TX data model (transactions, nullifiers, global state roots). It also adds support for parsing real Plonky2 and Groth16 proofs, using the crate from the original POC.

The main change is breaking the state management out of `Node` and into a dedicated `StateMachine`, which can be tested with mock transactions. A mock parser reads JSON-formatted "transactions" for use in tests without having to verify real proofs.

Also merges in `txlib`, with a four-element state root hash (Merkle roots of transaction hashes, nullifiers, past state roots, and current block number).


common/src/groth.rs
```diff
@@ -0,0 +1,119 @@
+//! Check out common/src/lib.rs documentation for context.
+//!
+
+use std::time::Instant;
+
+use anyhow::Result;
+use tracing::info;
+
+const INPUT_PATH: &str = "../tmp/plonky2-proof";
```
dhvanipa, line 9: Are the methods that use this path called server-side or client-side?

If it's client side, how does this path work with Tauri? is it inside the application folder created by the OS?

common/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+///     first compute the one extra recursive proof from the given MainPod's proof in
+///     order to shrink it, together with using the bn254's poseidon variant in the
+///     configuration of the plonky2 prover, in order to make it compatible with the
+///     Groth16 circuit.
+///     Then compute a Groth16 proof which verifies the last plonky2 proof
+#[cfg(feature = "groth16")]
+pub mod groth;
+pub mod payload;
+///   B) "shrink":
+///     first shrinks the given MainPod's proof, and then compresses it,
```
dhvanipa, line 15: in B), what does "compress" mean here? I assume its not to convert to groth16 as that's A)

common/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+}
+
+pub fn set_from_value(v: &Value) -> Result<containers::Set> {
+    match v.typed() {
+        pod2::middleware::TypedValue::Set(s) => Ok(s.clone()),
+        _ => Err(anyhow!("Invalid set")),
+    }
+}
+
+pub fn log_init() {
```
dhvanipa, line 86: can we follow the same logging convention we're already using in the synchronizer? Or update that one to use this

basically we should use the same logging everywhere in the repo

synchronizer/src/api.rs
```diff
@@ -6,29 +6,31 @@ use serde::Serialize;
...
 
 #[derive(Clone)]
 struct AppState {
-    node: Arc<Node>,
+    state_machine: Arc<StateMachine>,
 }
 
 #[derive(Serialize)]
 struct StateResponse {
     transactions: Vec<String>,
```
dhvanipa, line 19: Are `String`s fine here, or do we need to use `RawValue`?

txlib/src/examples/mine_ore.podlang
dhvanipa, line 1: Do we need the `examples` directory in this repo? I think having it will confuse things and make search harder, eg searching "pickaxe" would highlight this file even though it's not the application podlang

synchronizer/src/state_machine.rs
```diff
@@ -0,0 +1,487 @@
...
+                block_number, "Blob did not contain a valid TxFinalized proof; skipping"
+            );
+            return Ok(());
+        };
+
+        // A payload is only valid if it references a GSR we have previously computed.
+        // We accept any historical GSR, not just the latest, to tolerate blobs that were
+        // proven against a state root that has since been superseded.
+        {
+            let state = self
```
dhvanipa: We are reading and writing the state with this `map_err` in many places, it might be nice to have a helper function/macro

synchronizer/src/node.rs
```diff
@@ -185,10 +167,14 @@ impl Node {
...
         Self::log_slot_start(&slot_ctx);
-        self.log_current_state()?;
+        self.state_machine.log_current_state()?;
+
+        let block_number = slot_ctx.execution_block_number;
+
         if !slot_ctx.has_blob_commitments {
             debug!(slot = slot_ctx.slot, "Slot has no blob commitments");
-            return Ok(Some(slot_ctx.execution_block_number));
+            self.state_machine.advance_block(block_number as i64)?;
```
dhvanipa, line 176: By calling `advance_block` in all the places in this function, we are effectively producing a new global state root on every Ethereum block, correct?

synchronizer/.env.example
dhvanipa, line 1: I like how the env is under synchronizer. I think we can remove the top level `.env.example` then

justfile
dhvanipa, line 1: nice, ty for adding this

dhvanipa (approved):
nice work! I really like how you've made everything modular so it's easier to add tests & follows separation of concerns

common/src/groth.rs
robknight: This code was copied as-is from the POC repo - @ed255 might have more context on how it was used originally. I'm not using this particular code path here.

My understanding is that this would probably be used by the "relayer" as a proof compression step.

common/src/lib.rs
robknight: Again I'm not 100% sure here as I didn't write this.

common/src/lib.rs
robknight: Ok, I can update the logging here to match repo-wide convention.

synchronizer/src/api.rs
robknight: This is what will be returned by the server in the JSON, so this is the API response format. I don't see a problem since we have to convert the `RawValue`s to strings before returning them, and this data structure just makes the shape of the API response explicit.

txlib/src/examples/mine_ore.podlang
robknight: Good point, I'll remove it.

common/src/groth.rs
ed255: In the 2025 PoC this path was used in the client-side.
The treatment of this path and `OUTPUT_PATH` will not work great if used from an installed app; because it may write files in unexpected places (or fail to write due to permissions).
Even for a relayer it would be better to have a better solution for this; so that these paths are independent from the directory from which the process is started.

common/src/lib.rs
ed255: "compress" means calling this plonky2 method https://github.com/0xPARC/plonky2/blob/109d517d09c210ae4c2cee381d3e3fbc04aa3812/plonky2/src/plonk/proof.rs#L58
I don't know the details but reading the comments in plonky2 it seems this method mostly removes redundancy in the proof data structures (data that can be deterministically computed from the rest)

synchronizer/src/node.rs
robknight: Yes

common/src/lib.rs
robknight: On reflection, I'm not sure exactly what this means. Can you be specific about what you mean by the logging convention we're using the synchronizer?

common/src/lib.rs
dhvanipa: ahh plonky2 has a compress method which seems like standard compression. Could we update the comment so it refers to that?

common/src/lib.rs
dhvanipa: Here `log_init` is formatting logs with a specific time format, ie `.with(tracing_subscriber::fmt::layer().with_timer(timer))`. But in the synchronizer `tracing_subscriber` doesn't have this extra layer.

I think if we updated synchronizer main's method to call `log_init`, this would make it so there's only one place all the layers of logging live

common/src/groth.rs
dhvanipa: Since this code path is not used, we could add an issue to track this and not block on this PR being merged as it's a dependency for the re-org and relational DB PRs

common/src/payload.rs
```diff
@@ -0,0 +1,278 @@
...
+    Ok(elems)
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+#[allow(clippy::large_enum_variant)]
+pub struct Payload {
+    pub proof: PayloadProof,
+    /// Commitment of the finalized transaction dictionary `{live, nullifiers, state_root_hash}`.
+    pub tx_final: Hash,
+    pub state_root_hash: Hash,
```
ed255, line 42: To save some bytes, instead of sending the hash we could send the slot/block number (as u64), and the synchronizer then gets the state root hash from the number (which is a mapping that the synchronizer already has)

common/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+
+/// POD proving:
+/// 2 options to prepare the POD proofs:
+///   A) "groth":
+///     first compute the one extra recursive proof from the given MainPod's proof in
+///     order to shrink it, together with using the bn254's poseidon variant in the
+///     configuration of the plonky2 prover, in order to make it compatible with the
+///     Groth16 circuit.
+///     Then compute a Groth16 proof which verifies the last plonky2 proof
+#[cfg(feature = "groth16")]
```
ed255, line 11: This code supports two modes:
- "groth16", the plonky2 proof is verified by a groth16 circuit and the groth16 proof is sent on-chain
- "not(groth16)", the plonky2 proof is just shrunk and compressed and sent on-chain

In the MVP I don't think we need the two modes; we should pick one and remove the unused code.
To answer this we need to figure out some details:
- Do we want a relayer that accumulates DO commitments?
- What's the max number of commitments we want per transaction?

synchronizer/.env.example
```diff
@@ -0,0 +1,10 @@
+# Required
+RPC_URL=http://localhost:8545
+BEACON_URL=http://localhost:5052
+TO_ADDRESS=0x0000000000000000000000000000000000000000
```
ed255, line 4: Even if this is just an example, I propose using a different address, because this one is already used for other Ethereum applications; so we would be checking transactions that are unrelated to our application.  We should pick an address that nobody is using.

synchronizer/src/api.rs
```diff
@@ -6,29 +6,31 @@ use serde::Serialize;
...
+use crate::state_machine::StateMachine;
 
 #[derive(Clone)]
 struct AppState {
-    node: Arc<Node>,
+    state_machine: Arc<StateMachine>,
 }
 
 #[derive(Serialize)]
 struct StateResponse {
```
ed255, line 18: In pod2 many types already implement `Serialize + Deserialize`, so we could just use pod2 types here instead of manually converting types to strings.
For example we can have `nullifiers: Vec<RawValue>`.

Is there any reason for not following that approach?

synchronizer/src/db.rs
```diff
@@ -40,35 +44,49 @@ impl Db {
...
-    pub fn init(&self) -> Result<()> {
-        Ok(())
-    }
-
     pub fn load_state(&self) -> Result<DerivedState> {
         let mut transactions = HashSet::new();
         let mut nullifiers = HashSet::new();
+        let mut gsr_entries: Vec<(u64, Hash)> = Vec::new();
 
         for entry in self.db.iterator(IteratorMode::Start) {
```
ed255, line 52: I guess this will be replaced by a relational DB right?

synchronizer/src/api.rs
```diff
@@ -43,12 +45,22 @@ pub async fn run_api_server(
 async fn get_state(
     State(app_state): State<AppState>,
 ) -> Result<Json<StateResponse>, (StatusCode, String)> {
-    let (mut transactions, mut nullifiers) =
-        app_state.node.state_snapshot().map_err(internal_error)?;
+    let (transactions, nullifiers, global_state_roots) = app_state
```
ed255, line 48: I would strongly suggest not sending the full state via an API call, because this will keep growing and growing.
- This is not scalable
- This requires holding the entire state in memory

Instead I suggest having API endpoints that give the merkle proofs that the clients need to build transaction predicates.  This may uncover missing features in the pod2 library because we have always built statements with access to full trees; so the API for building statements with merkle proofs may be incomplete.

synchronizer/src/proof.rs
```diff
@@ -0,0 +1,244 @@
...
+    /// here so it does not need to be re-parsed on every blob.
+    pub fn new() -> Result<Self> {
+        let params = Params::default();
+        let module = txlib::predicates::module();
+        let txn_finalized_pred = module
+            .predicate_ref_by_name("TxFinalized")
+            .ok_or_else(|| anyhow!("TxFinalized not found in txlib module"))?;
+        let vds_root = DEFAULT_VD_SET.root();
+        let (common_circuit_data, verifier_circuit_data) =
+            &*cache_get_shrunk_main_pod_circuit_data(&params);
```
ed255, line 104: If we're not doing it already, I suggest warming up all the caches when the synchronizer starts.
This way once it starts processing all preprocessing has been done already and the first processed pod is representative of future processed pods in terms of timing.

synchronizer/src/state_machine.rs
```diff
@@ -0,0 +1,484 @@
...
+use anyhow::{anyhow, Result};
+use pod2::middleware::Hash;
+use tracing::{info, warn};
+
+use txlib::StateRoot;
+
+use crate::db::{Db, DerivedState, SyncProgress};
+use crate::proof::BlobParser;
+
+/// In-memory view of the consensus state, kept in sync with the database.
```
ed255, line 15: We will have to change this once we have support for persistent storage in MerkleTrees.  Once that lands we should avoid having big containers in memory.

synchronizer/src/db.rs
```diff
@@ -103,35 +121,64 @@ impl Db {
...
+    // Use block_number as u64 big-endian for lexicographic ordering
+    let block_u64 = block_number as u64;
+    [GSR_PREFIX, &block_u64.to_be_bytes()].concat()
 }
 
-fn tx_key(id: &str) -> Vec<u8> {
-    [TX_PREFIX, id.as_bytes()].concat()
+/// Serialize a Hash to 32 raw bytes for RocksDB storage.
+pub fn hash_to_db_bytes(hash: Hash) -> Vec<u8> {
+    let hex_str: String = hash.encode_hex();
```
ed255, line 170: I guess we'll refactor after we add the relational DB, but this is not nice: a hash is internally an array of bytes, there's no need to encode it as hex and decode it again.

synchronizer/src/state_machine.rs
```diff
@@ -0,0 +1,484 @@
...
+                    state.transactions.remove(&payload.tx_final);
+                    return Ok(());
+                }
+            }
+
+            self.db
+                .persist_transaction(payload.tx_final, slot, block_number)?;
+            for nullifier in &payload.nullifiers {
+                state.nullifiers.insert(*nullifier);
+                self.db.persist_nullifier(*nullifier, slot, block_number)?;
```
ed255, line 128: The list of nullifiers are written in a non-atomic batch, so the synchronizer could crash in the middle of processing a blob, and then restart with an inconsistent state: it would have missing nullifiers in the loaded state.
We should use atomic transactions to store things in the database.
Again, I guess this will be refactored with the relational DB.

ed255 (approved):
I've added a few comments, but I'm approving already because my comments are either:
- A topic of discussion that we can figure out later
- Issues in parts of the code that will be refactored with the relational DB

synchronizer/src/db.rs
dhvanipa: this db iterator is loading the txs, nullifier, and gsr's into memory. The relational DB is only for sync data, not for this derived data. I think this iterator goes away when we switch to storing these as merkle tree's on disk vs sets in memory.

synchronizer/src/db.rs
dhvanipa: I don't see a field in POD2 to get it as an array of bytes? Do you mean doing something like this?

```
pub fn hash_to_db_bytes(hash: Hash) -> Vec<u8> {
    let mut out = Vec::with_capacity(32);
    for limb in hash.0.iter().rev() {
        out.extend_from_slice(&limb.to_canonical_u64().to_be_bytes());
    }
    out
}
```
Though I think this is very similar to what `encode_hex` is already doing: https://github.com/0xPARC/pod2/blob/main/src/middleware/basetypes.rs#L129

ed255: sorry, you're right, there's no array of bytes; I was thinking about the array of field elements.
The snippet of code you wrote is what I was thinking about.  It can also be done just with iterators:
```rust
use plonky2::field::types::PrimeField64;
/// Serialize a Hash to 32 raw bytes for RocksDB storage.
pub fn hash_to_db_bytes(hash: Hash) -> Vec<u8> {
    hash.0
        .iter()
        .flat_map(|e| e.to_canonical_u64().to_le_bytes())
        .collect()
}
```

PR #8, opened 2026-03-03
Time

### LINKED ISSUE #6: Synchronizer should reject transactions citing a global state root from >300 blocks in the past
For supporting time predicates, we want a time resolution of one hour (300 blocks). The synchronizer should treat a transaction as invalid if its grounding global state root is greater than 300 blocks old.

robknight:
Closes #6.

Changes the synchronizer so that transactions referencing a global state root hash older than 300 blocks ago are rejected.

Implements predicates for two kinds of time:
- "Lock until", which locks an object until at least a certain number of blocks can be proven to have elapsed
- "Expiring option", which is an object or action that can only be used until a certain time.

The latter is expressed as a number of blocks, but this should be understood as having a coarse-grained resolution based on the maximum age enforced by the synchronizer. In other words, "expire after 1000 blocks" means "expire once the block `current_block + 1000` can no longer be used as the ground in a finalized transaction", which in practice means 1300 blocks later.

This might be confusing and may need some more thought or better documentation.


synchronizer/src/state_machine.rs
```diff
@@ -83,18 +92,30 @@ impl StateMachine {
...
-            if !state.global_state_roots.contains(&payload.state_root_hash) {
+            let Some(&gsr_block) = state.gsr_block_numbers.get(&payload.state_root_hash) else {
                 warn!(
                     slot,
                     block_number,
                     "Blob proof state_root_hash not found in known GSR history; rejecting"
                 );
                 return Ok(());
+            };
+            if let Some(current_block) = block_number {
```
dhvanipa: why is `block_number` an option in `process_blob`? Isn't it always defined

synchronizer/src/state_machine.rs
```diff
@@ -21,6 +25,9 @@ struct InnerState {
     /// Ordered history of Global State Roots, one per processed block.
     /// Blobs may reference any GSR in this history, not just the latest.
     global_state_roots: Vec<Hash>,
+    /// Maps each known GSR hash to the EL block number at which it was produced.
+    /// Used to enforce the maximum GSR age limit on incoming blobs.
+    gsr_block_numbers: HashMap<Hash, i64>,
```
dhvanipa, line 31: why is the block number signed? isn't it always positive?

timelib/src/predicates/time.podlang
```diff
@@ -0,0 +1,67 @@
...
+  DictContains(tx_before, "state_root_hash", state_root)
+  BlockNumberForStateRoot(state_root, block_number)
+  GtEq(timeout_block, block_number)
+  DictDelete(new_state, old_state, "timeout_block")
+  txlib::TxMutated(tx_after, tx_before, new_state, old_state)
+)
+
+UnlockObject(new_state, old_state, tx_after, tx_before,
+    private: state_root, prior_state_root, distance, old_tx) = AND(
+  DictContains(tx_before, "state_root_hash", state_root)
```
dhvanipa, line 49: why is this statement needed?

timelib/src/predicates/time.podlang
```diff
@@ -0,0 +1,67 @@
...
+  DictContains(state, "key", key)
+  DictContains(state, "value", value)
+  DictContains(state, "timeout_block", timeout_block)
+)
+
+// Exercises an expiring option in a single transaction. The grounding GSR's block
+// number must be <= timeout_block. The synchronizer enforces the grounding GSR is
+// at most one window (~300 blocks) old, so this is best understood as "times out
+// at block N" with ~300-block fuzz rather than a precise wall-clock deadline.
+ExecuteOption(new_state, old_state, tx_after, tx_before,
```
dhvanipa: I don't get this predicate. I think it'd be helpful to discuss it over a call

timelib/src/tx_utils.rs
```diff
@@ -0,0 +1,312 @@
...
+    let target = Value::from(gsr_when_locked.hash());
+    let idx = gsr_current
+        .gsrs
+        .array()
+        .iter()
+        .position(|v| v == &target)
+        .expect("gsr_when_locked not found in current state root's gsrs array")
+        as i64;
+
+    let distance = gsr_current.block_number - gsr_when_locked.block_number;
```
dhvanipa: might be nice to have a inline check here if distance > locked so can exit early if unlock is called too early

synchronizer/src/state_machine.rs
robknight: Good point, it is always defined and so doesn't need to be an option.

synchronizer/src/state_machine.rs
robknight: POD2 Ints are i64s, and we need to use the block number in PODs.

timelib/src/predicates/time.podlang
robknight: It's needed in order to extract the state root from `tx_before`, so that we can pass it to `DistanceBetweenStateRoots`. We can't (currently) use the anchored key syntax (`tx_before.state_root`) when passing arguments to custom predicates.

timelib/src/predicates/time.podlang
robknight: I've rewritten this and added a more comprehensive example.

timelib/src/predicates/time.podlang
```diff
@@ -0,0 +1,56 @@
...
+  BlockNumberForStateRoot(state_root, block_number)
+  BlockNumberForStateRoot(prior_state_root, prior_block_number)
+  SumOf(block_number, prior_block_number, distance)
+)
+
+// Proves the grounding GSR's block number is <= state.timeout_block. Assumes the
+// state dict contains a "timeout_block" field (the expirable-object mixin).
+// block_number is a public arg so callers can reuse it (e.g. SumOf for renewal).
+// Note: timeout_block has ~300-block fuzz due to the synchronizer's max-age window.
+NotExpired(state, block_number, tx_before,
```
dhvanipa, line 38: I think we should have a predicate that adds the `timeout_block` field as it's a specific field, vs it being inlined into the creature predicate. Similar to how we have lock and unlock

timelib/src/predicates/time.podlang
dhvanipa: the new example makes more sense, thank you!

I do think though that this example shouldnt be merged into this repo, for the same reason we didnt merge the txlib example. It can live in the 10object-exploration repo

robknight: I don't see how it is possible to have meaningful tests without a realistic example - otherwise we are testing the predicates in isolation but not for the kinds of use-cases we expect to have.

dhvanipa: We will have tests but they'll be for the actual predicates that the MVP is using vs sample ones (at least in this repo)

ie Wood and Creature are sample predicates that don't necessarily represent what we will ship. Once we have the predicates we're shipping, we should definitely have tests for those!

robknight: At that point we could replace these examples with more realistic examples, but in the meantime how are we supposed to know how the predicates are meant to be used?

dhvanipa: > but in the meantime how are we supposed to know how the predicates are meant to be used?

I agree we need examples (ie Creature) to know, but they shouldn't live in this repo. We create the predicates and test them with examples in 10items. Once we know they work, we move them here (without the examples), and then add the MVP predicates when that's ready

robknight: Does this mean that I should make two PRs, one in this repo and one in the 10 object repo, with the PR in this repo containing the code, and the PR in the 10 object repo containing the tests?

I am not sure how this is supposed to work in terms of CI, especially if this repo starts to depend on a newer version of POD2 than the 10 object repo does, or vice-versa.

dhvanipa: > Does this mean that I should make two PRs

yes! Features that are new and uncertain should start out in the 10 item repo, and then once tested there, moved here if we decide to include them in the MVP. eg this is what we did for txlib

Both repos can have different CIs, and we wouldnt port over a change to this repo if we're using a different version of POD2 here that doesnt support it (and for some reason we dont want to update it). 

eg when we move the SDK over, we'll bump POD in this repo to have the new params, but in the meantime this repo can have a different stable version of pod

experiment repo --> product repo

timelib/src/predicates/time.podlang
```diff
@@ -0,0 +1,56 @@
+use module 0xTXLIB_MODULE_HASH as txlib
+
+// Ensures that an object is locked for a minimum of `duration` blocks
+LockObject(new_state, old_state, tx_after, tx_before, duration) = AND(
+  DictInsert(new_state, old_state, "locked", duration)
+  txlib::TxMutated(tx_after, tx_before, new_state, old_state)
```
ed255: These predicates look good to me!
I have a small suggestion for optimization: Remove the `TxMutated` from `LockObject` and `UnlockObject` so that the user of the timelib can join those updates with others in a single `TxMutated`.

ed255 (approved):
I've mostly reviewed the predicate definitions (`time.podlang` and `creature.podlang`) and they both look good to me!
I think it will not be too hard to add add-on helpers in the SDK to use the time predicates.

timelib/src/examples/creature.podlang
```diff
@@ -0,0 +1,46 @@
...
+)
+
+// Feeds the creature, renewing its timeout_block. Proves the creature is valid
+// (IsCreature) and not yet expired. Verifies new_timeout = current_block + 300.
+Feed(state, state0, tx, tx0,
+    private: block_number, new_timeout) = AND(
+  IsCreature(state0)
+  time::NotExpired(state0, block_number, tx0)
+  SumOf(new_timeout, block_number, 300)
+  DictUpdate(state, state0, "timeout_block", new_timeout)
```
ed255: I noticed that here, if I don't know the details about `time::NotExpired` it's hard to understand the relation between that and updating `timeout_block`.

One idea that comes to my mind is having a predicate called `time::UpdateExpiration`; but that would be wasteful (unless we get custom predicate inlining)

Otherwise having some comment would help.  And for the SDK we can bundle all these together in a single helper function, which also helps.

PR #18, opened 2026-03-05
handle reorgs

dhvanipa:
## Reorg Handling

The synchronizer tracks canonical beacon roots per slot (`slot_root:<slot>`) and checks them while syncing.

Reorg is detected when any of these happen:

- a slot that previously had a root is now missing
- a slot that was previously processed as empty is now present with a block
- stored root for a slot differs from newly fetched root
- fetched header `parent_root` does not match stored root of the previous slot

When detected, it:

1. walks backward to find the last slot where stored root matches live root
2. rewinds DB state to just before divergence
3. replays forward from the divergence slot

Rollback removes:

- tx/nullifier/gsr entries seen after the retained slot
- stored slot-root entries after the retained slot
- sync progress after the retained slot

### DIFF
--- a/synchronizer/src/config.rs
+++ b/synchronizer/src/config.rs
@@ -19,6 +19,8 @@ pub struct AppConfig {
 }
 
 pub fn load_config() -> Result<AppConfig> {
+    let _ = dotenvy::from_filename("synchronizer/.env");
+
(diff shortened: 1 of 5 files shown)


robknight (approved):
lgtm!

PR #20, opened 2026-03-05
Add CI

### LINKED ISSUE #19: Add continuous integration
This repo lacks a GitHub Actions config.

robknight:
Closes #19 

These configs are copied from https://github.com/0xPARC/pod2-client. In that repo we also had JavaScript tests for the Tauri UI, and a "release" task which would build the Tauri app for various platforms (Linux, MacOS, Windows). We could copy that across when we want to start automated builds for distribution, but that seems unnecessary at this stage.

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -1,0 +1,25 @@
+name: Clippy Check
+
+on:
+  pull_request:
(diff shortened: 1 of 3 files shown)


dhvanipa (approved):
nice!

Issue #14, opened 2026-03-03
What should the synchronizer public API be?


robknight:
Currently there is an API for sending the entire state history, which is bad (see https://github.com/dobjlabs/zk-craft/pull/5#discussion_r2877526819).

PR #21, opened 2026-03-05
add relational db

dhvanipa:
## Summary
Refactors `synchronizer` storage into a split model:

- **`SyncDb` (Postgres)** is now the control plane for sync metadata:
  - sync cursor
  - canonical slot roots/status
  - per-slot apply journal for idempotent recovery
- **`AppDb` (RocksDB)** now stores **only app-derived state**:
  - transactions
  - nullifiers
  - GSR history

This removes sync metadata from RocksDB and moves reorg/cursor logic to relational state.

## What changed
- Added `sqlx` Postgres integration and new module: `sync_db.rs`.
- Renamed storage modules/types for clarity:
  - `db.rs` -> `app_db.rs`, `Db` -> `AppDb`
  - `sync_pg.rs` -> `sync_db.rs`, `SyncPg` -> `SyncDb`
- Updated sync runtime flow:
  1. derive slot delta purely from current state snapshot (no live in-memory mutation)
  2. persist slot as `pending` in Postgres + journal
  3. apply delta to RocksDB
  4. finalize slot as `applied` in Postgres and advance cursor
  5. apply the same delta to live in-memory state
- Added startup recovery for pending/unapplied slots via PG journal replay.
- Reorg rollback now:
  - rewinds canonical metadata in Postgres
  - uses returned journal deltas to delete affected app keys in RocksDB
- API `/state` progress fields now come from Postgres sync metadata.
- Added local DB auto-create for `SYNC_METADATA_DB`.
- Added explicit log for missing/empty beacon slots.
- Restored and adapted `state_machine` test coverage to delta-based API.

## Config/env updates
- New/renamed vars:
  - `APP_STATE_DB` (optional, default `data/synchronizer-db`)
  - `SYNC_METADATA_DB` (optional, default `postgres://postgres@localhost:5432/synchronizer`)
- Removed old `DB_PATH`.
- Updated `README` and `.env.example` accordingly.
- Added `just reset-local-state` to wipe `data/` and recreate local Postgres `synchronizer` DB.

## Why
- Keeps app state in KV (Merkle-friendly path) while using relational storage for canonical sync metadata/reorg handling.
- Improves crash recovery and idempotency with a journaled apply/finalize model.
- Avoids RocksDB-wide metadata rollback scans by using Postgres as sync authority.
- Prevents API/state drift by avoiding speculative in-memory mutation before durable finalize.


synchronizer/src/api.rs
dhvanipa, line 1: The API will eventually return merkle proofs vs the entire state from memory/postgress, but this is unused right now, so is for a future PR and I've not updated it here

Already being tracked here https://github.com/dobjlabs/zk-craft/issues/14

dhvanipa: actually to be less confusing I just removed the `/state` endpoint for now, so we dont need to update the code until we actually need it

synchronizer/src/sync_loop.rs
```diff
@@ -170,7 +243,9 @@ async fn find_divergence_slot(node: &Node, current_slot: u32) -> Result<u32> {
         }
         slot = prev_slot;
     }
-    Ok(0)
+    Err(anyhow!(
+        "No common ancestor found while handling reorg at slot {current_slot};"
```
robknight, line 247: If I understand correctly, this means that we are not able to handle a reorg which begins before the first block that we have synchronized. The old code treated `0` as a special case which effectively means "re-sync from the start", but that option no longer exists.

This might be a reasonable decision (maybe such deep reorgs are unlikely) but I want to know if this is intentional.

synchronizer/src/node.rs
```diff
@@ -157,19 +193,28 @@ impl Node {
...
-    ) -> Result<Option<u32>> {
+    ) -> Result<ProcessedSlot> {
         let Some(slot_ctx) = self.build_slot_context(beacon_block_header).await? else {
-            return Ok(None);
+            return Ok(ProcessedSlot {
+                slot: beacon_block_header.slot,
+                block_root: beacon_block_header.root,
+                parent_root: beacon_block_header.parent_root,
+                block_number: None,
+                is_empty: true,
```
robknight, line 215: Is this correct? If we got here it's because we expected to find a block, but if we didn't find one should we be throwing an error rather than saying the block is empty?

robknight (commented):
I think this looks good overall, but I have a couple of questions. Happy to approve once those are answered.

synchronizer/src/sync_loop.rs
dhvanipa: yup! It is indeed intentional.

I realized a bunch of complexity could be removed if we made this assumption. And in that error case, what we could do is wipe the DB and restart the process, vs having the code auto handle it.

synchronizer/src/node.rs
dhvanipa: We don't necessarily expect to find a block here. We have a valid slot, but not every consensus block is guaranteed to have an execution block. A slot can be skipped or empty, so it's not 1:1

robknight (approved):
lgtm!

PR #23, opened 2026-03-10
Port sdk from 10 objects

ed255:
This PR ports the sdk from the 10 objects repository (originally reviewed in https://github.com/0xPARC/10object-exploration/pull/14)

I improved the ergonomics of the sdk, you can review that change independently here: https://github.com/0xPARC/10object-exploration/compare/feat/decl-craft?expand=1#diff-823537328d87a0888e29996a5a5b0208d3fb78b294f9090dbcc62358b805e424R1160
Basically instead of using only structs and enum with native constructors I added methods so that the declaration can be made using the builder pattern.  My main motivation was to be able to mix `&'static str` and `String` and a nice way to do that is having a method that receives `impl Into<String>`.

I've also ported:
- vdf pod (no changes)
- pow pod (no changes)
- txlib and pod2utils small changes required by the sdk

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,9 +1,10 @@
 [workspace]
-members = ["app-gui/src-tauri", "common", "pod2utils", "synchronizer", "txlib"]
+members = ["app-gui/src-tauri", "common", "pod2utils", "synchronizer", "txlib", "craft_sdk", "intro_pods/vdfpod", "intro_pods/lt_eq_u256_pod"]
 resolver = "2"
 
 [workspace.dependencies]
-pod2 = { git = "https://github.com/0xPARC/pod2", rev = "cdf227e3530247e07e375daed007cae7d443b4ae", default-features = false, features = [
+# Branch `tmp/more-wildcards`
+pod2 = { git = "https://github.com/0xPARC/pod2", rev = "9993799c08e6e95b11eb016763fcc13f5030293e", default-features = false, features = [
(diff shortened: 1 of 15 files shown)


intro_pods/lt_eq_u256_pod/src/lib.rs
dhvanipa, line 1: nice, thanks for putting the intro pods under a folder vs top level!

Cargo.toml
```diff
@@ -1,9 +1,10 @@
 [workspace]
-members = ["app-gui/src-tauri", "common", "pod2utils", "synchronizer", "txlib"]
+members = ["app-gui/src-tauri", "common", "pod2utils", "synchronizer", "txlib", "craft_sdk", "intro_pods/vdfpod", "intro_pods/lt_eq_u256_pod"]
 resolver = "2"
 
 [workspace.dependencies]
-pod2 = { git = "https://github.com/0xPARC/pod2", rev = "cdf227e3530247e07e375daed007cae7d443b4ae", default-features = false, features = [
+# Branch `tmp/more-wildcards`
```
dhvanipa, line 16: nice ty for including a comment with the branch name!

I wonder if/when it would make sense to merge that parameter change into main?

dhvanipa (approved):
lgtm!

dhvanipa:
re the SDK change:
I like builder pattern as it shows up everywhere in Rust lol

PR #24, opened 2026-03-10
add cache to github actions

dhvanipa:
The current Github test action has become quite slow (~30 minutes). Based on the logs most of the time is spent in compiling the Rust crates. This PR adds a cache system so the previous build can be re-used. The cache is broken when a crate is added/modified/deleted.

Uses [https://github.com/Swatinem/rust-cache](https://github.com/Swatinem/rust-cache)


robknight (approved):

robknight:
This is great. Thanks for figuring this out!

PR #25, opened 2026-03-11
[FOR REFERENCE] End to End Workflow

dhvanipa:
## Summary
Add end-to-end desktop GUI workflow (inventory -> action -> proof -> relayer -> synchronizer -> local object updates), plus the backend services and shared plumbing needed to support it.

Demo: https://www.youtube.com/watch?v=Je58Qvwy0aA

## What changed
- Added a full Tauri-based `app-gui` frontend/backend stack.
- Added GUI panels and state management (`InventoryPanel`, `ActionGrid`, `ContextPanel`, `ProofRunnerPanel`, `SettingsModal`, Zustand `useStore`).
- Implemented Tauri command layer for inventory load, action execution, object IO, settings, CPU sampling, and global state root fetch.
- Implemented action run pipeline in GUI backend (`run_sdk_action`) with relayer submission + synchronizer confirmation.
- Added object watcher/events so GUI refreshes on `.dobj` changes.
- Added generated action/class ID typing flow (`gen_ids`) for frontend type safety.
- Added action/class hash support in UI payloads and hover/tooltips for class hashes.
- Updated proof runner to use a single polled global state root (rendered for both valid/nullified rows).
- Added a new `relayer` service crate (API, DB models, worker, config, Ethereum submission flow). See `relayer/README.md` for more info
- Expanded `synchronizer` APIs/types to support GUI/relayer integration.
- Moved shared proof/blob logic into `common` crate and updated dependents.
- Updated docs/tooling (`app-gui/README`, `justfile`, `mprocs.yaml`, env examples).

## Known Issues

Before this PR is merged in, I will turn these into Github issues for further discussion and to be resolved in future PRs.
- [ ] The proof artifacts are not preloaded so the first proof takes MUCH longer than subsequent ones
- [ ] The CPU time estimate for each action is hardcoded 
- [ ] The prototype UI shows a "Global Valid State Root" and "Global Nullified State Root", but in the synchronizer there is a single global state root. Currently, we're rendering the same root for both for this reason, but the UI needs to be updated to show only one
- [ ] The action and class hash calculated is wrong. It uses `predicate.hash()`, but this doesn't include the module id.
- [ ] If the user closes the desktop app/fails to get the commit response, then the object file is not created. However, it's already added to the state, so now the user doesn't know an object got created and an existing one nullified
- [ ] Relayer has no spam protection/auth, anyone can submit a proof and it posts it on chain
- [ ] There is no status in the client for synchronizer/relayer so it's not obvious if an action might fail because of that until they do an action.
- [ ] Synchronizer is serving the entire state to clients for proof generation. Once we support persistent merkle trees, only the relevant merkle proofs should be served.
- [ ] test to make sure it works on Windows

Future (potential) features:
- [ ] Onboarding (reveal UI step by step)
- [ ] Desktop notification when an action finishes
- [ ] Ability to do multiple actions in GUI, vs having to wait for one to finish
- [ ] Relayer batches multiple proofs into a single tx 
- [ ] Synchronizer public dashboard (can be linked from GUI)


wandcrafting:
Screenshot 2026-03-12 at 4 24 17 PM

can this be a toggle between "cpu panel" and "action execution"

wandcrafting:
Screenshot 2026-03-12 at 4 26 10 PM

can this be more descriptive even if its simulated/faked to suggest something about the internals of whats going on

like hashing
and then verifying each of the input objects

wandcrafting:
Screenshot 2026-03-12 at 4 30 04 PM

can we change "bind all inputs" to something else. maybe like provide inputs, or just the action button but greyed out

wandcrafting:
Screenshot 2026-03-12 at 4 42 26 PM

in the ui, we are labeling each arg with its class (or eventually an identifier)

but the sdk labels args with both name and a class (and eventually even an identifier)

so maybe in the ui, we should label each arg with the title being its name and the subtitle being its class (or eventually identifier). example of a title with a subtitle below it: 
Screenshot 2026-03-12 at 4 43 40 PM

hovering on the label should show the hash of the class:
Screenshot 2026-03-12 at 4 45 03 PM

wandcrafting:
Screenshot 2026-03-12 at 4 50 01 PM

we don't have any examples of objects here using Time yet, so this should be a tailend TODO

but in the object's key-values view, we should style and name the universal fields differently (lock, expiry) than the other fields

"    - For time-locks rate limits, the passage of time causes the object to become unlocked or available for further use. If the object has a lock which expires in one day, then anyone who wants to use it must prove that at least one day has elapsed since the lock was placed on the object. Similarly for a rate limit which says that an object can only be used once per hour, the user must prove that at least one hour has elapsed since the last use.
    - For expiry, the user has to prove something more complicated: that the option to use the object has not expired."

dhvanipa:
> Screenshot 2026-03-12 at 4 26 10 PM
> can this be more descriptive even if its simulated/faked to suggest something about the internals of whats going on
> 
> like hashing and then verifying each of the input objects

I think it makes more sense to verify the inputs when they're selected vs once you click the action.

Yea we can include a bunch of fake steps like what shows up in a loading screen

dhvanipa:
> Screenshot 2026-03-12 at 4 42 26 PM
> in the ui, we are labeling each arg with its class (or eventually an identifier)
> 
> but the sdk labels args with both name and a class (and eventually even an identifier)
> 
> so maybe in the ui, we should label each arg with the title being its name and the subtitle being its class (or eventually identifier). example of a title with a subtitle below it: Screenshot 2026-03-12 at 4 43 40 PM
> 
> hovering on the label should show the hash of the class: Screenshot 2026-03-12 at 4 45 03 PM

hovering on the label already shows the class, see video timestamp 7:00 and 7:04

image

image

can also add subtitle

wandcrafting:
> > Screenshot 2026-03-12 at 4 26 10 PM
> > can this be more descriptive even if its simulated/faked to suggest something about the internals of whats going on
> > like hashing and then verifying each of the input objects
> 
> I think it makes more sense to verify the inputs when they're selected vs once you click the action.
> 
> Yea we can include a bunch of fake steps like what shows up in a loading screen

Screenshot 2026-03-13 at 7 46 09 AM

1. generating recursive proof
steps: verifying existing proofs/dobjs and spending compute / hashing

wandcrafting:
> > Screenshot 2026-03-12 at 4 42 26 PM
> > in the ui, we are labeling each arg with its class (or eventually an identifier)
> > but the sdk labels args with both name and a class (and eventually even an identifier)
> > so maybe in the ui, we should label each arg with the title being its name and the subtitle being its class (or eventually identifier). example of a title with a subtitle below it: Screenshot 2026-03-12 at 4 43 40 PM
> > hovering on the label should show the hash of the class: Screenshot 2026-03-12 at 4 45 03 PM
> 
> hovering on the label already shows the class, see video timestamp 7:00 and 7:04
> 
> image image
> can also add subtitle

on hover: show hash/identifier (like right now)
title: class label or identifer (like row now)
subtitle: name (as per sdk)

wandcrafting:
we can get estimates for cpu time cost based on if there's a vdf/pow thing in a given action in the sdk

dhvanipa:
Merged in via https://github.com/dobjlabs/zk-craft/pull/28 and https://github.com/dobjlabs/zk-craft/pull/29

PR #28, opened 2026-03-13
Add Relayer & Update Synchronizer API

dhvanipa:
- Add the relayer service, see `relayer/README.md` for overview
- Updated synchronizer with 4 new endpoints needed by the client, see `synchronizer/README.md` for details


relayer/src/main.rs
```diff
@@ -0,0 +1,110 @@
...
+use eth::{EthClient, EthGateway};
+use worker::{run_worker, WorkerConfig};
+
+/// Boot relayer dependencies, then run API server and background worker together.
+#[tokio::main]
+async fn main() -> Result<()> {
+    common::log_init();
+
+    let cfg = load_config()?;
+    debug!(?cfg, "Loaded relayer config");
```
robknight: The config contains a private key, so depending on where logs get stored this might be a security concern, e.g. for users who have some kind of cloud log monitoring service.

relayer/src/api.rs
```diff
@@ -0,0 +1,469 @@
...
+            )
+                .into_response(),
+            ApiError::NotFound(msg) => (
+                StatusCode::NOT_FOUND,
+                Json(serde_json::json!({"error": msg})),
+            )
+                .into_response(),
+            ApiError::Internal(err) => (
+                StatusCode::INTERNAL_SERVER_ERROR,
+                Json(serde_json::json!({"error": err.to_string()})),
```
robknight: I wonder if this might leak internal data. It might be safer to log the error with a hash/nonce, and return the nonce to the client so that we can search the logs to find the error content.

relayer/src/worker.rs
```diff
@@ -0,0 +1,529 @@
...
+                now,
+                block_number,
+                format!("receipt timeout after {}s", timeout_secs),
+            )
+            .await?;
+            return Ok(());
+        }
+    }
+
+    job.attempt_count = job.attempt_count.saturating_add(1);
```
robknight: Each job has a single attempt count, but we use it for both the original submission and for polling, and we don't distinguish between polling failures which are due to network errors and failures which are due to the tx simply not being visible yet.

So we spend 1 attempt to submit the tx. Then if we spend 7 attempts polling, we exceed the maximum number of attempts. We can keep polling but now that we are over the maximum number of attempts, any polling request which hits a network failure will cause the whole job to be marked as failed, even if the orignal submission succeeded.

I'm not sure that we need to increment the job attempt count for polling at all, given that we have a separate `receipt_timeout_secs` mechanism. `attempt_count` could just be used for submission.

justfile
```diff
@@ -5,23 +5,37 @@
...
-    rm -rf data/
-    psql postgres://postgres@localhost:5432/postgres -c 'DROP DATABASE IF EXISTS synchronizer;'
+# Run all ignored test
+test-ignored:
+    cargo test --workspace --release -- --ignored --nocapture
 
-# Run the slow end-to-end proof test 
+# Run the slow end-to-end proof test
 test-e2e:
     cargo test -p synchronizer test_e2e_real_proof -- --ignored --nocapture
```
robknight: Tests which use real proofs should be run with `--release` (I know this isn't new in this PR but I only just noticed this)

relayer/src/api.rs
```diff
@@ -0,0 +1,469 @@
...
+
+    let payload = app_state
+        .parser
+        .parse_blob(&payload_bytes)
+        .map_err(|err| ApiError::BadRequest(format!("payload verification failed: {err}")))?
+        .ok_or_else(|| {
+            ApiError::BadRequest("payload did not decode into a valid proof payload".to_string())
+        })?;
+
+    let tx_final = payload.tx_final.encode_hex::<String>();
```
robknight: In the synchronizer we use the `encode_hash_hex` function which adds a `0x` prefix. This means that the tx hashes/state roots returned by the `/jobs/:id` API will not match those from the synchronizer's `/state` API, unless the code matching them is smart enough to treat both strings as equivalent.

robknight (changes requested):
Good overall, just some minor changes to consider.

relayer/src/main.rs
dhvanipa: Yes true, private key is leaked and so is the RPC API key. Removed the log both in synchronizer and relayer!

relayer/src/api.rs
dhvanipa: Yes true, changed it to return a unique id!

relayer/src/worker.rs
dhvanipa: Yes true, removed this increment!

justfile
dhvanipa: Updated!

relayer/src/api.rs
dhvanipa: Good catch! Updated it to also have the prefix!

dhvanipa:
> Good overall, just some minor changes to consider.

Addressed all changes!

robknight (approved):
lgtm!

PR #51, opened 2026-03-17
Add deploy instructions for synchronizer & relayer

dhvanipa:


.github/workflows/clippy.yml
dhvanipa, line 1: I made this change to align the other actions which are using release so they produce the same `target` artifact

robknight (approved):
Looks good to me, although I have not tested deploying it.

dhvanipa:
> Looks good to me, although I have not tested deploying it.

This was the exact script I ran on Friday to deploy it, so it's been tested there

PR #52, opened 2026-03-17
Create .dobj file before commit

### LINKED ISSUE #34: Create .dobj file before commit
If the user closes the desktop app/fails to get the commit response, then the object file is not created. However, it's already added to the state, so now the user doesn't know an object got created and an existing one nullified

dhvanipa:
Fixes https://github.com/dobjlabs/zk-craft/issues/34

This fixes the case where an action is committed, but the desktop app closes or never receives the final commit response. Previously, the state could already be updated while the local object files were never written, leaving the user unaware that a new object existed and an old one had been nullified.

The fix saves object-file changes before waiting for the relayer/synchronizer response, rolls those changes back if the commit actually fails, and reconciles local inventory against synchronizer state on load so missed objects/nullifications are recovered.

It also adds a grounded/pending state in the inventory UI so newly created objects are visible immediately (with yellow dot) but are not usable until the commit is observed:
image


robknight (approved):
lgtm!

PR #53, opened 2026-03-17
MCP server

robknight:
Adds an MCP server to the main app, and a proxy MCP client which connects to the main app via HTTP. The README includes instructions for configuring Claude to use the MCP server in conversations. Claude should be capable of figuring out what is going on, and performing crafting operations when connected to the app.

When compiling/restarting, sometimes it is necessary to restart Claude or to inspect Claude's logs in Settings -> Developer.

### DIFF
--- a/app-gui/src-tauri/src/lib.rs
+++ b/app-gui/src-tauri/src/lib.rs
@@ -15,7 +15,7 @@ use objects::{
 use sdk::{get_global_state_root, load_gui_inventory, run_sdk_action, ActionRunGate};
 use settings::{build_app_menu, get_app_settings, handle_settings_menu_event, save_app_settings};
 
-const MCP_PORT: u16 = 3001;
+use craft_mcp::DEFAULT_PORT as MCP_PORT;
 
 #[cfg_attr(mobile, tauri::mobile_entry_point)]
 pub fn run() {
@@ -67,8 +67,8 @@ async fn start_mcp_server(app: tauri::AppHandle) -> Result<(), Box<dyn std::erro
     let app_settings = get_app_settings(app.clone())?;
 
     let ops = mcp::AppCraftOps::new(objects_dir, app, app_settings);
-    let config = zk_craft_mcp::McpConfig::default();
-    let server = zk_craft_mcp::McpServer::new(ops, config);
+    let config = craft_mcp::McpConfig::default();
+    let server = craft_mcp::McpServer::new(ops, config);
 
     let listener = tokio::net::TcpListener::bind(format!("127.0.0.1:{MCP_PORT}")).await?;
     eprintln!("zk-craft: MCP server listening on http://127.0.0.1:{MCP_PORT}/mcp");
--- a/app-gui/src-tauri/src/lib.rs
+++ b/app-gui/src-tauri/src/lib.rs
@@ -1,9 +1,12 @@
 mod cpu;
 mod error;
+mod mcp;
 mod objects;
 mod sdk;
 mod settings;
 mod spec;
+
+use std::sync::Arc;
 
 use cpu::{sample_app_cpu, CpuMonitor};
 use objects::{
@@ -12,11 +15,14 @@ use objects::{
 use sdk::{get_global_state_root, load_gui_inventory, run_sdk_action, ActionRunGate};
 use settings::{build_app_menu, get_app_settings, handle_settings_menu_event, save_app_settings};
 
+const MCP_PORT: u16 = 3001;
+
 #[cfg_attr(mobile, tauri::mobile_entry_point)]
 pub fn run() {
     if let Err(err) = common::load_dotenv() {
         eprintln!("zk-craft: failed to load app-gui env: {err}");
     }
+
     tauri::Builder::default()
         .menu(build_app_menu)
         .on_menu_event(|app, event| {
@@ -24,11 +30,20 @@ pub fn run() {
         })
         .plugin(tauri_plugin_opener::init())
         .manage(CpuMonitor::new())
-        .manage(ActionRunGate::new())
+        .manage(Arc::new(ActionRunGate::new()))
         .setup(|app| {
             if let Err(err) = start_objects_watcher(app.handle().clone()) {
                 eprintln!("zk-craft: objects watcher disabled: {err}");
             }
+
+            // Start MCP server
+            let handle = app.handle().clone();
+            tauri::async_runtime::spawn(async move {
+                if let Err(err) = start_mcp_server(handle).await {
+                    eprintln!("zk-craft: MCP server failed: {err}");
+                }
+            });
+
             Ok(())
         })
         .invoke_handler(tauri::generate_handler![
@@ -45,4 +60,19 @@ pub fn run() {
         ])
         .run(tauri::generate_context!())
         .expect("error while running tauri application");
+}
+
+async fn start_mcp_server(app: tauri::AppHandle) -> Result<(), Box<dyn std::error::Error>> {
+    let objects_dir = objects::objects_dir(&app)?;
+    let app_settings = get_app_settings(app.clone())?;
+
+    let ops = mcp::AppCraftOps::new(objects_dir, app, app_settings);
+    let config = zk_craft_mcp::McpConfig::default();
+    let server = zk_craft_mcp::McpServer::new(ops, config);
+
+    let listener = tokio::net::TcpListener::bind(format!("127.0.0.1:{MCP_PORT}")).await?;
+    eprintln!("zk-craft: MCP server listening on http://127.0.0.1:{MCP_PORT}/mcp");
+
+    server.serve(listener).await?;
+    Ok(())
 }
--- a/app-gui/src/App.tsx
+++ b/app-gui/src/App.tsx
@@ -10,6 +10,7 @@ import {
   listenOpenSettings,
   listenObjectsChanged,
   listenRunSdkActionProgress,
+  listenMcpActionStarted,
   openObjectsDir,
   sampleAppCpu,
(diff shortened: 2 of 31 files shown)


mcp/src/bin/mock_server.rs
```diff
@@ -0,0 +1,31 @@
+use tracing_subscriber::{EnvFilter, layer::SubscriberExt, util::SubscriberInitExt};
+use zk_craft_mcp::mock::MockCraftOps;
+use zk_craft_mcp::{McpConfig, McpServer};
+
+const BIND_ADDRESS: &str = "127.0.0.1:3001";
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    tracing_subscriber::registry()
```
dhvanipa: this should be replaced with `common::log_init();`

mcp/src/bin/proxy_stdio.rs
```diff
@@ -0,0 +1,107 @@
+/// Stdio-to-HTTP proxy for the ZK-Craft MCP server.
+///
+/// Claude Desktop launches this as a child process. It reads JSON-RPC from
+/// stdin, forwards to the Tauri app's streamable HTTP MCP endpoint, and
+/// writes responses to stdout.
+use std::io::{BufRead, Write};
+
+use tracing_subscriber::{EnvFilter, layer::SubscriberExt, util::SubscriberInitExt};
+
+const DEFAULT_URL: &str = "http://127.0.0.1:3001/mcp";
```
dhvanipa: Two different places define this string, it might be better to store it in one place, at least the base url with the port

mcp/src/mock.rs
dhvanipa, line 1: Is this mock just for tests?

mcp/src/types.rs
```diff
@@ -0,0 +1,157 @@
...
+    pub produced_by: Vec<String>,
+    /// Actions that consume objects of this class
+    pub consumed_by: Vec<String>,
+}
+
+// -- Actions --
+
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct RunActionInput {
```
dhvanipa, line 115: Some of these types are duplicate definitions from what already exists, can they be imported instead? eg `RunSdkActionInput`

mcp/Cargo.toml
```diff
@@ -0,0 +1,43 @@
+[package]
+name = "zk-craft-mcp"
```
dhvanipa: I feel like the package name shouldnt have "zk-craft" as a prefix as none of the other packages in this repo do

mcp/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+An MCP (Model Context Protocol) server that exposes ZK-Craft's digital object operations to AI agents. Claude Desktop or Claude Code can inspect inventory, explore crafting actions, read documentation, and execute ZK proof-based crafting operations through this server.
+
+## Architecture
+
+```
+Claude Desktop <--stdio--> zk-craft-mcp-proxy <--HTTP--> Tauri app (MCP + GUI)
+```
+
+The Tauri app embeds a streamable HTTP MCP server on port 3001. A thin stdio proxy binary bridges Claude Desktop (which expects stdio transport) to the app's HTTP endpoint. The GUI and MCP server share in-process state — when Claude runs an action, the GUI shows real-time progress.
```
dhvanipa, line 11: is this proxy pattern common? Why doesn't the app just serve stdio directly?

app-gui/src/App.tsx
```diff
@@ -99,6 +100,41 @@ function App() {
     };
   }, [applyRunSdkActionProgress]);
 
+  // Listen for MCP-initiated actions so the proof panel shows progress
+  useEffect(() => {
+    let cancelled = false;
+    let unlisten: (() => void) | null = null;
+    import("@tauri-apps/api/event")
```
dhvanipa: instead of importing it in-line like this, it'd be better if this listen method was in `app-gui/src/shared/api/tauriClient.ts`, where the other listen methods are already

app-gui/src/shared/state/store.ts
```diff
@@ -245,13 +250,7 @@ export const useStore = create<AppState>((set) => ({
...
       };
     }),
-  runProof: async ({ actionId, inputBindings, cpuCost }) => {
-    const postDoneHoldMs = 2800;
-    const verifyTargets =
-      inputBindings.length > 0
-        ? inputBindings.map((binding) => binding.label)
-        : ["(no inputs)"];
-
+  initProofPanel: ({ actionId, cpuCost, args }) =>
```
dhvanipa, line 253: Why was `runProof` renamed to `initProofPanel`? Isnt it still running the proof?

app-gui/src-tauri/src/lib.rs
```diff
@@ -1,33 +1,48 @@
...
+use std::sync::Arc;
+
 use cpu::{sample_app_cpu, CpuMonitor};
 use objects::{
     get_objects_dir, open_objects_dir, pick_dobj_file_path, read_dobj_file, start_objects_watcher,
 };
 use sdk::{get_global_state_root, load_gui_inventory, run_sdk_action, ActionRunGate};
 use settings::{build_app_menu, get_app_settings, handle_settings_menu_event, save_app_settings};
 
+const MCP_PORT: u16 = 3001;
```
dhvanipa: related to my other comment before, if this is defined in the `mcp` package, then it can be easily imported here. Then, we only need to change one place to update the port

app-gui/src-tauri/src/lib.rs
```diff
@@ -45,3 +60,18 @@ pub fn run() {
...
+
+async fn start_mcp_server(app: tauri::AppHandle) -> Result<(), Box<dyn std::error::Error>> {
+    let objects_dir = objects::objects_dir(&app)?;
+    let app_settings = get_app_settings(app.clone())?;
+
+    let ops = mcp::AppCraftOps::new(objects_dir, app, app_settings);
+    let config = zk_craft_mcp::McpConfig::default();
+    let server = zk_craft_mcp::McpServer::new(ops, config);
+
+    let listener = tokio::net::TcpListener::bind(format!("127.0.0.1:{MCP_PORT}")).await?;
```
dhvanipa, line 73: same comment for URL being defined in one place for MCP as port

mcp/src/server.rs
```diff
@@ -0,0 +1,497 @@
...
+                    })
+                    .unwrap_or_else(|| "Resource not found".to_string())
+            }
+        }
+    }
+}
+
+// -- Instructions --
+
+const INSTRUCTIONS: &str = "\
```
dhvanipa: could this live in a separate `.md` file? so its easy to modify the prompt?

mcp/src/mock.rs
dhvanipa: nvm it seems like it's for the mock MCP server which is used for testing MCP without running the app

mcp/src/bin/mock_stdio.rs
```diff
@@ -0,0 +1,22 @@
+use rmcp::ServiceExt;
+use rmcp::transport::io::stdio;
+use tracing_subscriber::{EnvFilter, layer::SubscriberExt, util::SubscriberInitExt};
+use zk_craft_mcp::mock::MockCraftOps;
+use zk_craft_mcp::server::CraftMcpService;
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    // Log to stderr so stdout stays clean for JSON-RPC
```
dhvanipa: I think it'd be good to also move this to common with a new method, eg `log_stderr` (or an argument on the existing `log_init`)

mcp/README.md
```diff
@@ -0,0 +1,160 @@
...
+      "command": "/absolute/path/to/zk-craft/target/release/zk-craft-mcp-proxy",
+      "args": ["--port", "3001"]
+    }
+  }
+}
+```
+
+### 3. Start the Tauri app
+
+The proxy connects to the app's MCP endpoint. The app must be running first:
```
dhvanipa: The app alone relies on the synchronize and relayer, so maybe this should actually be instructing the user to do `just dev` at the root

dhvanipa (approved):
small comments on organization, but otherwise lgtm!

Been trying it out and it's great, it made me a diagram:

Image

dhvanipa (changes requested):
oh actually the action buttons in the GUI are no longer working on this branch..

mcp/README.md
robknight: stdio only works if the app is a child process of Claude, which can't be the case for the main GUI app. The proxy application runs as a child process and then communicates with the main app over HTTP.

app-gui/src/shared/state/store.ts
robknight: In the GUI, we want to show the panel and trigger the running of the proof. But if we initiated running the proof from an MCP command, we need a way to cause the panel to show which doesn't run the proof again, so these were separated out. This allows the GUI to call `runProof` which also opens the panel, while allowing the MCP server to trigger opening the panel via an event while running the proof via a different mechanism.

mcp/src/types.rs
robknight: These are intentionally separate - the MCP types derive `JsonSchema` and are shaped for LLM consumption, while the `app-gui` types are used for Tauri IPC. The `mcp` crate has no dependency on `app-gui` (the `CraftOps` trait is the boundary), so sharing them would mean either a circular dependency or a new shared crate just for two identical-looking structs. The explicit conversion in mcp.rs keeps the two layers decoupled.

dhvanipa (approved):
lgtm!

PR #54, opened 2026-03-18
Add release action

dhvanipa:
Adapted from https://github.com/0xPARC/pod2-client/blob/main/.github/workflows/release.yml

This PR adds a release pipeline for the Tauri desktop app and the packaging changes needed to make the native SCIP runtime ship correctly with release artifacts. 

The Apple build is signed & notarized by my personal Apple Developer account (ad-hoc signed version was not working due to the SCIP libraries not being signed). 

I don’t love how much complexity now lives in `app-gui/src-tauri/build.rs`, but it exists because the packaged app needs native SCIP libraries staged differently per platform, and macOS in particular requires extra install-name and signing work for the app bundle to launch cleanly and pass notarization. Without that logic, the release bundle either misses native dependencies or picks them up in a way that breaks runtime loading/signing.

One future alternative, if feasible, is to statically link or otherwise remove the external SCIP runtime dependency entirely.

Sample run: https://github.com/dobjlabs/zk-craft/releases/tag/untagged-4b9cea3ab9815f741da9


app-gui/src-tauri/examples/gen_ids.rs
dhvanipa, line 1: I moved this to examples so it doesn't get included in the release binary (previously there was app-gui and gen_id binaries)

robknight (approved):
I tested the aarch64 MacOS release and it worked for me.

PR #56, opened 2026-03-18
refactor for the new pod2 value api for persistent containers

ed255:
I've also changed all String errors in tauri-src to `anyhow::Error` for better ergonomics, this way we can just return any error from functions without manually turning them into strings.  For the `tauri::command` I added a new error that can be serialized and that implements `From<anyhow::Error>`

The changes related to the new Value API look like this:
- Casting a Value.
  - Before: `i64::try_from(value.typed_value())...`
  - After: `value.as_int()`
- Getting the raw from a Value (I think this change could have been done even before)
  - Before `RawValue::from(value)`
  - After: `value.as_raw()`
- Removed all references of `TypedValue` in favor of just `Value`

### DIFF
--- a/app-gui/src-tauri/src/error.rs
+++ b/app-gui/src-tauri/src/error.rs
@@ -1,0 +1,16 @@
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum CommandError {
+    #[error(transparent)]
+    Anyhow(#[from] anyhow::Error),
+}
+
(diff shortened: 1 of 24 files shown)


app-gui/src-tauri/src/error.rs
dhvanipa, line 1: hmm there is already `src/shared/error.ts` so maybe this could be consolidated with that

dhvanipa (approved):

PR #60, opened 2026-03-23
Use persistent storage for synchronizer app state

### LINKED ISSUE #37: update synchronizer to serve merkle proofs instead of full state

dhvanipa:
This PR refactors the synchronizer to persist canonical app state as POD2-backed Merkle containers instead of materializing full transaction/nullifier/GSR sets in memory.

It makes Postgres the sole source of the canonical head, storing the canonical roots and head metadata as normalized columns on each canonical slot, while RocksDB becomes only the content-addressed backing store for persistent POD2 Merkle nodes and values. It also removes the old staged apply/rollback journaling flow and replaces the old full-state API with compact head, membership, nullifier, and grounding-witness endpoints.

On the client side, txlib, craft_sdk, and the Tauri app now ground actions from committed roots plus synchronizer-supplied Merkle proofs rather than downloading the full state.


dhvanipa:
This is `cargo test -p timelib predicates::tests::prove_lock_and_unlock -- --nocapture` failing sometimes with:
```
thread 'predicates::tests::prove_lock_and_unlock' (1502257) panicked at timelib/src/predicates/mod.rs:263:18:
called `Result::unwrap()` on an `Err` value: Frontend(Inner: Raw(0xcec693b56bc31b7d415d81501ed1b8bd3006f307e2ed98b012df78966609f31d) not a container
disabled backtrace)
```

Not sure if this is an accurate diagnosis but 2 guesses from different AIs:

### Wildcards:

A wildcard that is used both (eg `old_tx.live` in [`UnlockObject`](https://github.com/dobjlabs/zk-craft/blob/main/timelib/src/predicates/time.podlang#L47)):
1. as a full container, and
2. as the root of an anchored key

can end up bound as only the raw commitment hash instead of the typed container value.

Later, when a statement needs that same wildcard as a container again, proving fails with: `Frontend(Inner: Raw(0x...) not a container`

This seems to happen when a predicate is large enough to be split/reordered by apply_predicate. The failure is intermittent: if the wildcard gets bound first from `old_tx.live`, it is stored as a raw value; if it gets bound first from `old_tx`, the proof succeeds.

### add_entries_contains:

When a `SetContainsFromEntries` (or `ArrayContainsFromEntries`) operation
produces a `Contains` statement with a raw hash root, and that statement is
later passed as an argument to a custom predicate operation **in a different
MainPodBuilder** (as happens during `build_single_pod` in MultiPod proving),
`add_entries_contains` matches on `Statement::Contains(Literal, Literal,
Literal)` and auto-generates a `DictContains` operation for it. The
auto-generated `DictContains` has `OperationAux::None`, so `fill_in_aux`
calls `as_container_or_err` on the raw root value — which fails with
`"<hash> not a container"`.

The bug is masked when both operations live in the same MainPodBuilder
because `insert()` records the Contains key in `dict_contains`, preventing
`add_entries_contains` from generating a duplicate. In MultiPod proving,
`build_single_pod` creates a *fresh* MainPodBuilder for each POD, so when
the SetContains and the custom predicate land in different PODs (which
depends on the LP solver), `dict_contains` is empty and the bug triggers.

The reason it's flaky is whether the LP solver puts the SetContains operation and the custom predicate operation in the same POD or different PODs. When they're in the same MainPodBuilder, insert() pre-populates dict_contains and the spurious generation is skipped. When they're in different PODs (fresh builder, empty dict_contains), the spurious DictContains is generated and fails.

Suggested fix: Skip generating `DictContains` when the root value is not a container:

ed255:
> This is `cargo test -p timelib predicates::tests::prove_lock_and_unlock -- --nocapture` failing sometimes with:
> 
> ```
> thread 'predicates::tests::prove_lock_and_unlock' (1502257) panicked at timelib/src/predicates/mod.rs:263:18:
> called `Result::unwrap()` on an `Err` value: Frontend(Inner: Raw(0xcec693b56bc31b7d415d81501ed1b8bd3006f307e2ed98b012df78966609f31d) not a container
> disabled backtrace)
> ```

I'm investigating this.

I've been able to make the test deterministic by using a sequence of keys in the objects instead of random keys; in that case the test always fails.
Following the guesses by the AI the following could be happening:
- Randomized keys lead to different sorting of statements by the multi pod solver (this is my explanation for the random failure).
- For certain orderings a Dict contains ends up in some intermediate pod with RawValue for root and no proof in Aux (this is the AI guess).

ed255:
I believe I've fixed the issue, it was coming from pod2 and it was coming from the behavior of the MainPodBuilder (it had some bugs in edge cases); and while fixing it I've also simplified the MultiPod solver.  The fix is not yet clean for a PR but I've made this branch that you can use for this PR https://github.com/0xPARC/pod2/tree/next/2026-03-24

Tomorrow I'll continue working on the fix: cleaning it up and making a PR for pod2.  For today you can point the pod2 dependency in zk-craft to the `next/2026-03-24` to avoid getting blocked or annoyed by this issue.

dhvanipa:
> I believe I've fixed the issue, it was coming from pod2 and it was coming from the behavior of the MainPodBuilder (it had some bugs in edge cases); and while fixing it I've also simplified the MultiPod solver. The fix is not yet clean for a PR but I've made this branch that you can use for this PR https://github.com/0xPARC/pod2/tree/next/2026-03-24
> 
> Tomorrow I'll continue working on the fix: cleaning it up and making a PR for pod2. For today you can point the pod2 dependency in zk-craft to the `next/2026-03-24` to avoid getting blocked or annoyed by this issue.

Thank you! Not blocked by the issue, this PR is complete and ready for review. Just found it annoying in Github CI and felt like a bug, so wanted to bring it up

txlib/src/lib.rs
```diff
@@ -231,11 +261,27 @@ impl TxBuilder {
...
+                    OperationAux::MerkleProof(source_tx_proof),
+                ))
+                .unwrap();
+            let st_tx_in_state_root = ctx
+                .builder
+                .priv_op(Operation::custom(
+                    tx_in_state_root_pred.clone(),
+                    [st_state_root, st_tx_membership],
+                ))
+                .unwrap();
```
ed255: I recommend not creating custom statements manually because the `st_custom!` macro handles many things for you, including constructing predicates that have been split.
```rust
            let st_tx_in_state_root = st_custom!(
                ctx,
                TxInStateRoot() = (st_state_root, st_tx_membership)
            )
            .unwrap();
```

txlib/src/lib.rs
```diff
@@ -389,18 +435,86 @@ pub fn new_obj() -> Dictionary {
...
+        transactions: HashSet<Hash>,
+        nullifiers: HashSet<Hash>,
+        gsrs: Vec<Hash>,
+    }
+
+    fn test_hash(byte: u8) -> Hash {
+        Hash::from_hex(hex::encode([byte; 32])).expect("valid test hash")
+    }
+
+    impl TestState {
```
ed255: I feel like I've seen this code a few times already.  Could you unify all implementations of all cases and put them in a common crate to reuse it in tests?

ed255: Similar code exists in:
- timelib/src/predicates/mod.rs `tests`
- craft_sdk/src/lib.rs `tests`

synchronizer/src/app_db.rs
```diff
@@ -1,231 +1,348 @@
...
+}
+
+/// Compact committed app-state snapshot stored in RocksDB under `meta/head`.
+///
+/// The actual transaction/nullifier/GSR contents live in persistent POD2 containers;
+/// `AppHead` stores the roots and counts used to reopen those containers and serve
+/// state/proof queries.
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct AppHead {
```
ed255: I don't understand why this data is stored both in RocksDB and Postgress.

In my mental model this only needs to be stored in Postgress indexed by slot and the process of synchronizing a slot would go like this:
- Update any of the trees stored in RocksDB
- Save their roots in a new AppHead in Postgress

When querying the trees:
- Load the roots of a particular slot from Postgress
- Use those roots to access the merkle trees stored in RocksDB

synchronizer/src/sync_db.rs
```diff
@@ -220,101 +264,99 @@ impl SyncDb {
         }
     }
 
-    /// Stage canonical slot metadata and journal atomically as `pending`.
+    /// Stage a slot's canonical metadata and head-transition journal atomically as `pending`.
```
ed255: I don't understand this journaling strategy.
Isn't it enough to insert all rows related to new processed slots in a single atomic transaction that is commited after the RocksDB has been updated with the new trees?

And for reorgs just remove all rows for invalidated slots and restart from the biggest slot in the DB, without touching RocksDB at all?

synchronizer/src/app_db.rs
dhvanipa: yes you are right, the current model has a BUNCH of complexity from the previous design that can be greatly simplified. I will refactor it

synchronizer/src/sync_db.rs
dhvanipa: Same comment as [above](https://github.com/dobjlabs/zk-craft/pull/60#discussion_r2989282269), I will refactor

txlib/src/lib.rs
dhvanipa: updated to use macro!

txlib/src/lib.rs
dhvanipa: refactored into common crate!

dhvanipa:
I've refactored the synchronizer to be much simpler now, it's ready for a review again @ed255 @robknight

app-gui/src-tauri/src/sdk/synchronizer_client.rs
```diff
@@ -38,89 +46,122 @@ pub(crate) fn fetch_synchronizer_state(sync_api_url: &str) -> Result<Synchronize
...
     }
 
-    let payload: TxContainsResponse = response
-        .json()
-        .map_err(|err| anyhow!("failed to decode synchronizer tx/contains response: {err}"))?;
-    let mut present = HashSet::new();
-    for entry in payload.results {
+    let mut source_tx_proofs = HashMap::new();
+    let mut missing = Vec::new();
+    for entry in payload.source_tx_proofs {
```
robknight: Here we check that all of the hashes in the response have `present = true`, but we don't check that the response contains all of the hashes that were in the request - in other words, we are not validating that the response contains everything that the client expects. I think it would be good to validate this assumption here, even if the current version of the synchronizer always returns a response for each hash in the request.

synchronizer/src/sync_db.rs
```diff
@@ -71,13 +54,11 @@ impl SyncDb {
...
     ///
-    /// Creates:
-    /// 1. `sync_cursor` for single-row progress tracking
-    /// 2. `canonical_slots` for canonical slot metadata + status
-    /// 3. `slot_apply_journal` for per-slot apply/rollback replay
-    /// 4. supporting indexes for recovery and lookup paths
+    /// `canonical_slots` stores per-slot metadata plus the committed canonical roots and metadata
+    /// as regular SQL columns.
+    /// `sync_cursor` points at the latest canonical slot that should be treated as current.
     async fn bootstrap(&self) -> Result<()> {
```
robknight, line 61: I presume that this is intended to be used with a fresh DB/synchronizer instance and not as an upgrade/migration of an existing synchronizer?

app-gui/src-tauri/src/sdk/run_action.rs
```diff
@@ -389,14 +360,15 @@ pub(crate) async fn run_sdk_action_core(
...
-
     let resolved_inputs = resolve_inputs(&input, &descriptor)?;
-
-    verify_inputs_grounded(&app_settings.synchronizer_api_url, &resolved_inputs)?;
+    let source_tx_hashes = resolved_inputs
+        .iter()
+        .map(|input| input.record.spendable().tx.dict().commitment())
+        .collect::<Vec<_>>();
+    let grounding_witness =
+        fetch_grounding_witness(&app_settings.synchronizer_api_url, &source_tx_hashes)?;
```
robknight: `run_sdk_action_core` is async, but `fetch_grounding_witness` creates a blocking request. I think the right thing to do here is `spawn_blocking` and await.

synchronizer/src/api.rs
```diff
@@ -104,123 +146,248 @@ async fn get_state_head(
...
-
-    let snapshot = app_state
-        .state_machine
-        .api_state_snapshot()
+async fn post_grounding_witness(
+    State(app_state): State<AppState>,
+    Json(body): Json<GroundingWitnessRequest>,
+) -> Result<Json<GroundingWitnessResponse>, (StatusCode, String)> {
+    let snapshot = load_current_snapshot(&app_state).await?;
+    let source_tx_hashes = body
```
robknight, line 236: I think we might want a limit on the number of hashes here, to prevent DoS attacks.

synchronizer/src/api.rs
```diff
@@ -104,123 +146,248 @@ async fn get_state_head(
...
-        })
-        .collect();
+    Json(body): Json<MembershipRequest>,
+) -> Result<Json<MembershipResponse>, (StatusCode, String)> {
+    let MembershipContext {
+        head,
+        tx_hashes,
+        nullifiers,
+        membership,
+    } = load_membership_context(&app_state, &body.tx_hashes, &body.nullifiers).await?;
```
robknight, line 198: Similar concern about the unbounded number of hashes here, it is probably a good idea to enforce a limit (same for other user query inputs).

synchronizer/src/sync_db.rs
```diff
@@ -58,7 +37,11 @@ pub struct SyncDb {
...
 
 impl SyncDb {
-    /// Connect to Postgres, ensure database exists, and bootstrap schema/indexes.
+    /// Connect to the synchronizer's Postgres metadata database.
+    ///
+    /// Postgres is the sole source of canonical heads. Each committed slot stores its
+    /// `CanonicalHead`,
+    /// while RocksDB only stores the content-addressed Merkle node/value backing store.
     pub async fn connect(database_url: &str) -> Result<Self> {
         ensure_database_exists(database_url).await?;
```
robknight, line 46: This is convenient but in production we probably don't want the regular DB user to have DB creation privileges. Probably not worth fixing here but we could log this as something to come back to later.

robknight (changes requested):
Looks good overall, just a few small tweaks that I think would make this more robust. For the robustness/DoS issues it might also be OK to create follow-up issues, so long as we get to them before serious third-party use begins.

app-gui/src-tauri/src/sdk/synchronizer_client.rs
dhvanipa: updated!

synchronizer/src/sync_db.rs
dhvanipa: yes

app-gui/src-tauri/src/sdk/run_action.rs
dhvanipa: updated

synchronizer/src/api.rs
dhvanipa: added!

synchronizer/src/api.rs
dhvanipa: added!

synchronizer/src/sync_db.rs
dhvanipa: opened issue: https://github.com/dobjlabs/zk-craft/issues/61

robknight (approved):
lgtm!

synchronizer/src/app_db.rs
```diff
@@ -1,231 +1,231 @@
...
-        let global_state_roots = gsr_entries.into_iter().map(|(_, h)| h).collect();
+        Self::open(db_path).with_context(|| format!("Failed to open RocksDB at path {db_path}"))
+    }
 
-        Ok(DerivedState {
-            transactions,
-            nullifiers,
-            global_state_roots,
-            gsr_block_numbers,
+    fn open(path: impl AsRef<Path>) -> Result<Self> {
```
ed255: `open` and `connect` seem redundant?

synchronizer/src/sync_db.rs
```diff
@@ -1,55 +1,34 @@
...
-/// `last_processed_block_number` is auxiliary execution-layer progress metadata.
+/// Current canonical head plus cursor metadata loaded from Postgres.
 #[derive(Debug, Clone, Copy)]
-pub struct SyncProgress {
-    pub last_processed_slot: u32,
+pub struct CurrentSnapshot {
+    pub head: CanonicalHead,
+    pub last_processed_slot: Option<u32>,
     pub last_processed_block_number: Option<u32>,
 }
```
ed255, line 21: If my understanding is correct, many types that the synchronizer has contain `Option<T>` which is used when the synchronizer has not yet synchronized an initial state of the Digital Objects, is this correct?

If so, I think the design could be simplified by removing all those options and assuming a genesis state that happens at a particular slot and block number for a particular CanonicalHead (which sould be just all subtrees empty).  Then on database creation rows for that genesis slot would be inserted.

synchronizer/src/sync_db.rs
```diff
@@ -71,13 +54,11 @@ impl SyncDb {
         Ok(store)
     }
 
-    /// Create schema objects used by the sync control plane.
+    /// Create the Postgres schema used by the synchronizer control plane.
```
ed255, line 57: What's the advantage of having `sync_cursor` vs just using `canonical_slots` for the same purpose?
For example, to get the last processed slot you could query `canonical_slots` for the row wight highest `slot`.

synchronizer/src/sync_db.rs
```diff
@@ -220,102 +251,70 @@ impl SyncDb {
...
+                head_nullifiers_root,
+                head_state_root_gsrs_root,
+                head_gsr_history_root,
+                head_current_gsr,
+                head_current_block_number,
+                head_tx_count,
+                head_nullifier_count,
+                head_gsr_count
+            )
+            VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15)
```
ed255, line 234: why would there be a conflict here?  Doesn't the synchronizer remove the rows after a reorg before re-inserting?

synchronizer/src/api_types.rs
```diff
@@ -1,58 +1,164 @@
...
     pub current_gsr: Option<String>,
+    /// Per-hash membership results.
     pub results: Vec<TxContainsEntry>,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
+/// Batch nullifier-membership request.
+pub struct NullifierContainsRequest {
+    /// Nullifier hashes to look up in the canonical nullifiers set.
+    pub nullifiers: Vec<String>,
```
ed255, line 70: I think the code could be simplified if you just use `RawValue`, or `Hash` for all these `String` cases; and rely on the Serialize/Deserialize impl of `RawValue`/`Hash` implemented in pod2.
Or is there a reason to choose a different serialization here?

The benefits of this is: less transformation on `api.rs` (both for inputs and outputs), but also we can reuse these types in the client and directly get the types we want, without an extra step of turning a `String` into `RawValue`/`Hash`.

ed255 (commented):
Overall looks good but I've left a few suggestions/discussions about parts that I believe can be simplified.  Perhaps we could discuss the details in a call to accelerate this.

synchronizer/src/app_db.rs
dhvanipa: yes, it can be just one

synchronizer/src/sync_db.rs
dhvanipa: yup true, `sync_cursor` is not needed anymore since we're storing head info in `canonical_slots`

synchronizer/src/sync_db.rs
dhvanipa: yea it was defensive, but will actually mask an unexpected duplicate-commit bug, so it should be removed

ed255 (approved):

synchronizer/src/api_types.rs
dhvanipa: Added issue to track: https://github.com/dobjlabs/zk-craft/issues/63

PR #64, opened 2026-03-30
Add re-try logic for RPC methods

dhvanipa:
My local synchronizer kept crashing due to instability in the RPC. Restarting it fixes it. This PR adds re-try logic instead so the process doesn't need to be restarted


robknight (approved):
lgtm!

PR #65, opened 2026-04-01
parallel batch fetch for synchronizer catch-up

dhvanipa:
- When the synchronizer is behind the beacon head by more than `batch_size` slots, fetch headers and blocks in parallel batches instead of one-at-a-time, significantly reducing catch-up time
- Batch size and sync delay defaults are derived from the RPC rate limit (15 req/s) to stay within API limits automatically; both remain overridable via `CATCHUP_BATCH_SIZE` and `SYNC_DELAY_MS` env vars

### DIFF
--- a/synchronizer/src/config.rs
+++ b/synchronizer/src/config.rs
@@ -9,6 +9,7 @@ const DEFAULT_HTTP_BIND: &str = "127.0.0.1:3000";
 const DEFAULT_SYNC_DELAY_MS: u64 = 333;
 const DEFAULT_RPC_RETRIES: u32 = 6;
 const DEFAULT_RPC_RETRY_MS: u64 = 1_000;
+const DEFAULT_CATCHUP_BATCH_SIZE: usize = 16;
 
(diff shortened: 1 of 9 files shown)


synchronizer/src/config.rs
```diff
@@ -33,10 +49,35 @@ pub fn load_config() -> Result<AppConfig> {
...
+                "SYNC_DELAY_MS overridden via env; ensure this respects the RPC rate limit"
+            );
+            ms
+        }
+        Err(_) => default_sync_delay_ms(),
+    };
+
+    let catchup_batch_size = match dotenvy::var("CATCHUP_BATCH_SIZE") {
+        Ok(v) => {
+            let size: usize = v.parse()?;
```
robknight, line 69: It might be worth verifying that CATCHUP_BATCH_SIZE is > 0

robknight (commented):
lgtm, one small comment which is not very important.

robknight (approved):

PR #67, opened 2026-04-03
add driver crate

dhvanipa:
In preparation for a "driver program", this PR moves out the code that would be in a "driver" but is currently tied to the desktop app into its own crate

Tested crafting the entire tree manually and via MCP and still works


driver/src/clients/synchronizer_client.rs
```diff
@@ -0,0 +1,363 @@
...
+        timeout_secs: u64,
+        poll_interval_ms: u64,
+    ) -> Result<SynchronizerHead>;
+}
+
+#[derive(Debug, Default)]
+pub struct HttpSynchronizerClient;
+
+impl SynchronizerClient for HttpSynchronizerClient {
+    fn fetch_head(&self, sync_api_url: &str) -> Result<SynchronizerHead> {
```
ed255, line 58: wouldn't it be better for all network methods to be async?  Won't this block the UI?

driver/src/object_store.rs
```diff
@@ -0,0 +1,281 @@
...
+            .find(|entry| entry.file_name == *file_name)
+            .ok_or_else(|| anyhow!("object file not found: {file_name}")),
+        ObjectSelector::ObjectId(object_id) => entries
+            .iter()
+            .find(|entry| entry.record.id == *object_id)
+            .ok_or_else(|| anyhow!("object not found: {object_id}")),
+    }
+}
+
+pub(crate) fn matches_query(entry: &ObjectFileEntry, query: &ObjectQuery) -> bool {
```
ed255, line 168: this may not be required now, but perhaps at some point we could consider using an sqlite database to index all objects and classes.

driver/src/driver.rs
```diff
@@ -0,0 +1,459 @@
...
+            catalog: Arc::new(BuiltinActionCatalog::new()),
+            synchronizer: Arc::new(HttpSynchronizerClient),
+            relayer: Arc::new(HttpRelayerClient),
+            payload_builder: Arc::new(DefaultPayloadBuilder),
+        }
+    }
+}
+
+#[derive(Clone)]
+pub struct Driver {
```
ed255, line 76: This implementation is very stateless.  The `Driver` only contains "configuration" and teach method reads/writes from disk and/or does something with `deps`.

Did you consider a design where `Driver` is more stateful?  For example it could hold a vector of objects so that `list_objects` doesn't read from disk each time.

ed255 (approved):
I've taken a quick look to understand the architecture of the driver.  I've left a few questions but otherwise looks good to me.

driver/src/clients/synchronizer_client.rs
dhvanipa: hmm actually after reviewing the code again, it makes sense to keep these sync because the driver is designed to be synchronous

It won't block the UI, because the Tauri commands already push the whole driver call onto a blocking worker thread (since there's other heavier operations like the proof generation)

driver/src/driver.rs
dhvanipa: mentioned on call but writing it here as well: I think loading all the objects into state is not scalable, and so that should always be some disk call with some metadata in memory. There could be benefits from having other stuff in memory like recent synchronizer fetches

PR #71, opened 2026-04-06
match object structure v0 doc

dhvanipa:
Update to match the object structure v0 doc: https://www.notion.so/Object-Structure-v0-33a5d7149de480aea435dc7830e69aa3

- Replace the binary `nullifier`/`live` fields on `.dobj` files with a proper `ObjectStatus` enum (unknown → pending → live → nullified) and an optional `txHash` field tracking on-chain inclusion
- Rewrite the execute flow to update object status incrementally through each phase (save as unknown, mark pending on relayer confirmation, mark live on synchronizer confirmation) instead of rolling back on post-acceptance failures
- Remove `rollback_results` — on pre-acceptance relayer failure, output files are preserved as `unknown` so users can retry submission without regenerating proofs
- Move local storage from `~/.objects` to `~/.dobj/objects` and consolidate settings under `~/.dobj/`
- Update `sync_inventory` reconciliation to use the full status model: three-pass reconcile now handles nullification, restoration of failed nullifications, and promotion to live
- Update GUI, MCP layer, and mock server to use status/txHash instead of nullifier/live
- Removed legacy field sourceAction (not used anywhere)

### DIFF
--- a/driver/src/driver.rs
+++ b/driver/src/driver.rs
@@ -144,16 +144,15 @@ impl Driver {
             .iter()
             .map(|entry| entry.record.spendable().tx.dict().commitment())
             .collect::<HashSet<_>>();
-        let live_nullifiers = entries
+        let all_nullifiers = entries
             .iter()
-            .filter(|entry| !entry.record.is_nullified())
             .filter_map(|entry| object_nullifier_hash(&entry.record.obj).ok())
             .collect::<HashSet<_>>();
 
         let membership = self.deps.synchronizer.fetch_membership_with_nullifiers(
             &settings.synchronizer_api_url,
             &source_tx_hashes.iter().copied().collect::<Vec<_>>(),
-            &live_nullifiers.iter().copied().collect::<Vec<_>>(),
+            &all_nullifiers.iter().copied().collect::<Vec<_>>(),
         )?;
 
         reconcile_objects(
--- a/driver/src/driver.rs
+++ b/driver/src/driver.rs
@@ -367,25 +367,31 @@ impl Driver {
         };
 
         // Past this point the proof has been accepted by the relayer and may
-        // land on-chain at any moment. The output
-        // files stay as `unknown` and the inputs stay nullified. The next
-        // `sync_inventory` call will reconcile once the tx is observed.
+        // land on-chain at any moment. Output files stay as Unknown until the
+        // relayer broadcasts and we have a tx_hash.
         let waiting_label = format!("Waiting for relayer job {}", submit_response.job_id);
         reporter.on_step(ExecutionPhase::Commit, &waiting_label, &commit_ctx);
-        let confirmation = self.deps.relayer.wait_for_confirmation(
+
+        // Poll until the relayer has broadcast the tx and we have a tx_hash,
+        // then mark output files as Pending.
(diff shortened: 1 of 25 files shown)


driver/src/driver.rs
```diff
@@ -338,65 +345,100 @@ impl Driver {
...
-            }
+        // Confirmation returned an Ethereum tx hash — update outputs to Pending.
+        let eth_tx_hash = confirmation.tx_hash.as_deref();
+        if let Err(err) = update_output_files(
+            &self.paths,
+            &saved.output_files,
+            ObjectStatus::Pending,
+            eth_tx_hash,
+        ) {
+            eprintln!("zk-craft: failed to update output status to pending: {err}");
```
ed255: is it OK to not handle this error?  This could come from a file system error right?  Perhaps it's better to return it?

driver/src/execute.rs
```diff
@@ -48,6 +44,60 @@ pub(crate) fn reconcile_objects(
...
+        };
+        if on_chain_nullifiers.contains(&nullifier_hash) {
+            continue; // Confirmed nullified on-chain, keep as is.
+        }
+        let restored_record = StoredObjectRecord {
+            status: ObjectStatus::Unknown,
+            ..entry.record.clone()
+        };
+        if let Err(err) = write_object_file(paths, &restored_record, &entry.file_name) {
+            eprintln!(
```
ed255: There are more unhandled errors here.

driver/src/execute.rs
```diff
@@ -48,6 +44,60 @@ pub(crate) fn reconcile_objects(
...
+        let source_tx_hash = entry.record.tx.dict().commitment();
+        if !grounded_txs.contains(&source_tx_hash) {
+            continue;
+        }
+        let live_record = StoredObjectRecord {
+            status: ObjectStatus::Live,
+            ..entry.record.clone()
+        };
+        if let Err(err) = write_object_file(paths, &live_record, &entry.file_name) {
+            eprintln!(
```
ed255: and here

driver/src/execute.rs
```diff
@@ -189,35 +223,24 @@ pub(crate) fn save_results(
     })
 }
 
-pub(crate) fn rollback_results(
+/// Update the status and tx_hash of previously saved output files on disk.
+pub(crate) fn update_output_files(
```
ed255, line 230: Could this be a disk intensive operation?  This includes the plonky2 proof right?  so each time it's reading and writing more than 200 KiB of data.
We should pay attention here in the future.

ed255 (approved):
I've left a few comments but in general looks good to me

driver/src/driver.rs
dhvanipa: good catch, I agree it's better to return, updated!

driver/src/execute.rs
dhvanipa: updated to return

driver/src/execute.rs
dhvanipa: updated to return

driver/src/execute.rs
dhvanipa: yea true, something to pay attention and potentially find a way to only write the status

PR #72, opened 2026-04-07
update driver API

dhvanipa:
match https://www.notion.so/Driver-Design-Doc-33a5d7149de48058bfb1d34d5aa6d50f

- ObjectQuery.live (bool) → ObjectQuery.status (ObjectStatus) for filtering by any lifecycle status
- Remove ObjectDetail type; read_object/read_object_file now return ObjectSummary
- Remove ActionRunGate to allow parallel action execution
- Update app-gui MCP bridge to fetch predicate_source from get_class() instead
- Delete runtime.rs


ed255 (approved):
LGTM!

PR #75, opened 2026-04-10
Compute VDF and LtEqU256 verifier data hashes at runtime

dhvanipa:
Replaces hardcoded hex hash strings with hashes computed from the circuit's verifier data at init time, so they stay correct if the circuits change.

I was using a local pod2 build with different parameters and facing this issue

### DIFF
--- a/craft_sdk/src/lib.rs
+++ b/craft_sdk/src/lib.rs
@@ -1100,7 +1100,6 @@ mod tests {
     use std::sync::Arc;
 
     use common::test_state::TestState;
-    use hex::FromHex;
     use lt_eq_u256_pod::LtEqU256Pod;
     use pod2::{
         frontend::{MainPod, Operation},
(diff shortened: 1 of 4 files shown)


ed255 (approved):
LGTM because it solves the issue, but it introduces another issue:
Now each time that you run a test that uses those hashes, even if it's mock it will build the circuits which takes some time.

I suggest opening an issue to address that by caching the results using the cache system defined in pod2 which supports caching to disk.  That way the data will be loaded from disk except for the first time.

dhvanipa:
> I suggest opening an issue to address that by caching the results using the cache system defined in pod2 which supports caching to disk.

Is this cache in this repo or in pod2?

EDIT: nvm I think its in this repo

dhvanipa:
> LGTM because it solves the issue, but it introduces another issue: Now each time that you run a test that uses those hashes, even if it's mock it will build the circuits which takes some time.
> 
> I suggest opening an issue to address that by caching the results using the cache system defined in pod2 which supports caching to disk. That way the data will be loaded from disk except for the first time.

Updated, ready for review again

ed255 (approved):
LGTM!

PR #76, opened 2026-04-10
Add fee-bump (replace-by-fee) support for blob transactions

dhvanipa:
Blob TXs can get stuck in the mempool when the base fee rises after submission. The relayer now detects stale pending transactions and resubmits replacements at the same nonce with higher fees.

I was seeing this a lot locally when my objects would remain pending for minutes at times.

**Relayer changes:**
- New config: `FEE_BUMP_AFTER_SECS`, `FEE_BUMP_MULTIPLIER_PCT`, `FEE_BUMP_MAX` (disabled by default)
- Tracks nonce and bump history (`nonce`, `bump_count`, `prev_tx_hashes`) on relay jobs
- Fetches the original TX's fee caps from the node and uses `max(network, original)` as the base before applying the multiplier — prevents both "underpriced" rejections and "insufficient funds" from inflated blob fees
- When a TX disappears from the mempool, checks if the nonce was consumed and scans `prev_tx_hashes` to resolve the actual mined TX hash
- New API endpoint `GET /api/v1/proofs/by-tx-final/{tx_final}` for tx hash lookups
- Validates `FEE_BUMP_AFTER_SECS >= RECEIPT_POLL_SECS` at startup

**Driver changes:**
- After `wait_for_confirmation`, updates `.dobj` files with the confirmed tx hash if it differs from the initial one (fee bump occurred)
- `sync_inventory` resolves correct tx hashes from the relayer for pending objects

Tested locally by setting `FEE_BUMP_AFTER_SECS` to 10 so it always bumps the fee


relayer/src/worker.rs
```diff
@@ -228,6 +248,101 @@ async fn poll_submitted_job(
...
+                                            {
+                                                if receipt.success {
+                                                    mined_hash = Some(prev_hash.clone());
+                                                    mined_block = receipt.block_number;
+                                                    break;
+                                                }
+                                            }
+                                        }
+
+                                        if let Some(hash) = &mined_hash {
```
robknight: Above we scan the previous hashes, but the loop does not seem to guarantee that we will find one, so we have to use `if let` here. What happens if there's no hash? Should we still be setting the job status as confirmed?

relayer/src/eth.rs
```diff
@@ -116,6 +154,79 @@ impl EthGateway for EthClient {
         Ok(tx_hash)
     }
 
+    async fn get_next_nonce(&self) -> Result<u64> {
+        let count = self.provider.get_transaction_count(self.from).await?;
```
robknight: This may need to be `self.provider.get_transaction_count(self.from).pending().await?` to account for pending transactions.

relayer/src/worker.rs
robknight: In calls to `submit_payload_with_fees`, we pass our stored nonce in to ensure that it is used, but `submit_payload` does not take a nonce parameter. I think it would make sense for `submit_payload` to take a nonce parameter in the same way that `submit_payload_with_fees` does.

robknight (changes requested):
Good overall, just the `pending` change in `get_nonce` seems like it might be important. Aside from that, the other points are just questions/suggestions but not blockers.

relayer/src/worker.rs
dhvanipa: ahh good catch! We should only mark job status confirmed when we actually find a receipt. Updated it to keep polling

relayer/src/eth.rs
dhvanipa: Updated to use pending

relayer/src/worker.rs
dhvanipa: Good idea, updated to use `nonce`

robknight (approved):

PR #77, opened 2026-04-15
Reimplement SDK with a scripting engine

ed255:
The new sdk is implemented in the `sdk` workspace member.
I've left the previous sdk in `craft_sdk` so that `app-gui/src-tauri/` still builds with the intention of doing the migration in a later PR.

The sdk is not feature complete but it has a nice structure and supports enough features to work for a rich example.  The main feature missing compared to the previous sdk is support for dependencies (which is required for mining stone).

Please read the `sdk/README.md` for more details on the design and implementation.


dhvanipa (approved):
lgtm! A bunch of stuff will need to be completed which is todo and not integrated, but it is known this is WIP, and merging it in will allow for this branch to not fall too far behind, without impacting other changes as it's a separate crate

sdk/src/lib.rs
```diff
@@ -0,0 +1,1393 @@
...
+    };
+    let v = match v.try_cast_result::<Ref>() {
+        Ok(v) => return Ok(v),
+        Err(v) => v,
+    };
+    let v = match v.try_cast_result::<ArgHandle>() {
+        Ok(v) => return Ok(v.arg),
+        Err(v) => v,
+    };
+    _ = v;
```
dhvanipa, line 707: what is the point of doing `_ = v`? Also the error after this reads it againt

sdk/src/lib.rs
```diff
@@ -0,0 +1,1393 @@
...
+    };
+    let v = match v.try_cast_result::<Ref>() {
+        Ok(v) => return Ok(v.borrow().as_value().clone()),
+        Err(v) => v,
+    };
+    let v = match v.try_cast_result::<ArgHandle>() {
+        Ok(v) => return Ok(v.arg.borrow().as_value().clone()),
+        Err(v) => v,
+    };
+    _ = v;
```
dhvanipa, line 727: Same comment as above here

sdk/src/lib.rs
dhvanipa, line 1: this file is doing a lot, and imo would be better to split it up into separate files in the future

sdk/src/lib.rs
ed255: `v` is `Dynamic` and `try_cast_result` https://docs.rs/rhai/1.24.0/rhai/struct.Dynamic.html#method.try_cast_result
will try to cast the underlying type to the requested type while consuming the value.  If the cast is not possible, it returns the original value so that you can try again.
So this `v` doesn't have any error information, it's just the same v as the argument to `try_ref_from_dynamic`.

The intention of `_ = v;` was to signal that this last `v` is not used, so that the each case in the try cast looks the same.  But now that you mention it, I'm not sure if this is a standard pattern in Rust; and I just noticed that the returned error uses `v`, so there's no need for this line I think.

sdk/src/lib.rs
ed255: Yes I agree!  I tried to move some obvious parts to files (like `fmt_podlang.rs`)
Splitting this file requires some thought: for example it would be interesting to have each phase related code in a different file, but both phases share many structs!

PR #78, opened 2026-04-15
add groth16 proof backend support

dhvanipa:
We want to support both `plonky2` and `groth16` proof types as the relayer may use the groth16 proof type to pack more data, but users should be allowed to submit plonky2 proofs since groth16 requires ~12 GB RAM which not all users may have.

- Adds Groth16 as an optional proof backend alongside the default Plonky2, reducing on-chain blob payload size
- Introduces a proof_type setting (`"plonky2" | "groth16"`) plumbed through DriverSettings, the PayloadBuilder trait, and the GUI settings modal
- Implements Groth16 proof generation (`build_groth16_relayer_payload`) and verification (`verify_groth16_proof`) using a `[proof_len | proof | public_inputs]` framing format
- Moves Groth16 trusted-setup artifacts from hardcoded `../tmp/` to platform cache dirs (`~/.cache/pod2-groth16/`)
- Bumps `pod2_onchain` rev to main commit and adds `just groth16-setup` for one-time trusted setup


common/src/groth.rs
```diff
@@ -52,21 +67,26 @@ mod tests {
...
-            println!("generating groth16's trusted setup");
-            let result = pod2_onchain::trusted_setup(INPUT_PATH, OUTPUT_PATH);
+        let vk_marker = Path::new(&output_path).join("verifying.key");
+        if !vk_marker.exists() {
+            println!("generating groth16's trusted setup at {output_path}");
+            let result = pod2_onchain::trusted_setup(&input_path, &output_path);
             println!("trusted_setup result: {result}");
         } else {
-            println!("trusted setup already exists, skipping generation");
+            println!("trusted setup already exists at {output_path}, skipping");
```
ed255, line 89: All these prints should be cleaned up and replaced by debug or info logs

common/src/groth.rs
```diff
@@ -3,21 +3,36 @@
...
+        .map_err(|e| anyhow!("failed to create cache dir {}: {e}", path.display()))?;
+    Ok(path.to_string_lossy().into_owned())
+}
 
 /// initializes the groth16 prover memory, loading the artifacts. This method
 /// must be called before the `prove` method.
 pub fn init() -> Result<()> {
-    pod2_onchain::init(INPUT_PATH, OUTPUT_PATH)?;
+    let input = cache_path(CACHE_SUBDIR_INPUT)?;
+    let output = cache_path(CACHE_SUBDIR_OUTPUT)?;
```
ed255, line 28: The PoC used a single directory for
- a) circuit artifacts & verifying key
- b) plonky2 proofs to be compressed
- c) groth16 proofs generated

For a non-PoC solution like zk-craft I think `a` should use a cache directory (like this PR does) but `b, c` should use a temporary directory, because files in `b, c` aren't reused.

common/src/proof.rs
```diff
@@ -119,25 +134,61 @@ impl ProofParser {
...
+            PayloadProof::Groth16(framed) => self.verify_groth16_proof(&framed, public_inputs),
+            #[cfg(not(feature = "groth16"))]
+            PayloadProof::Groth16(_) => Err(anyhow!(
+                "Groth16 proof received but 'groth16' feature is not enabled"
+            )),
+        }
+    }
+
+    #[cfg(feature = "groth16")]
+    fn verify_groth16_proof(&self, framed: &[u8], _public_inputs: Vec<F>) -> Result<()> {
```
ed255, line 163: Ignoring the public inputs here is a very bad idea.
Notice that the public inputs in `verify_shrunk_main_pod` are calculated with the statement, which is what the commitment layer uses to update the state!
The public inputs used to verify the groth16 proof need to be derived from the public inputs calculated in `verify_shrunk_main_pod` instead of using alternative public inputs serialized after the proof.

driver/src/driver.rs
```diff
@@ -35,6 +35,7 @@ use crate::types::{
 pub trait PayloadBuilder: Send + Sync {
     fn build_payload(
         &self,
+        proof_type: &str,
```
ed255, line 38: what about using a boolean here?  like `groth16: bool`, this way you don't need to document the possible options, you don't have to handle invalid cases, etc.
Alternatively I would propose to define an enum with the two cases.

ed255 (changes requested):
I'm not approving yet because I think the handling of public inputs for the groth16 proof is unsafe

dhvanipa:
closing this for now as its not scoped to do for this release

PR #79, opened 2026-04-16
replace builtin actions with .pexe plugin system

dhvanipa:
Replace the hardcoded `craft_sdk` + `BuiltinActionCatalog` with a dynamic plugin system. Actions now ship as `.pexe` archives (zip of `manifest.toml` + `plugin.rhai`) that the driver scans from `~/.dobj/actions/` at startup.

- New `pexe` crate: archive format library + packaging CLI (`build`, `dump`, `--install`)
- New `driver/src/pexe_catalog.rs`: unpacks each `.pexe`, compiles its Rhai script through `sdk::Sdk`, and aggregates the action/class catalog
- New `plugins/craft-basics/`: FindLog, CraftWood, CraftSticks, CraftWoodPick, UseWoodPick migrated from Rust to a Rhai script
- Deleted: `craft_sdk/`, `driver/src/builtin.rs`, compile-time `gen_ids` machinery, legacy `cpu_cost` and `reads_block` metadata fields
- `driver::paths` refactor: `DriverPaths::from_dobj_root` as single source of truth
- SDK: `pow_obj_grind` now takes a full u256 Raw target; added `top_limb_u256` helper and proper `u256_gt`, was failing without it
- GUI: consumes class/action metadata dynamically instead of generated constants
- justfile: `pack-plugins` / `install-plugins` recipes


driver/src/pexe_catalog.rs
```diff
@@ -0,0 +1,368 @@
...
+            }
+            combined_podlang.push_str(&format!(
+                "// plugin: {}\n{}",
+                plugin.manifest.plugin.name, podlang_src
+            ));
+            plugin.podlang_src = podlang_src;
+
+            // Capture class metadata from the manifest.
+            for class in &plugin.manifest.classes {
+                if !class_meta.contains_key(&class.name) {
```
ed255, line 97: within a module class names are unique; but when you have multiple modules you may have different classes with the same name.
Here you're discarding classes from other modules that have the same name of already processed modules.
We need to figure out a way to fix this.

When thinking about collections of modules, we should support classes with the same name defined in different modules.  We can distinguish such cases by the module hash.
UI-wise we also need a solution:
- For example, if we have collisions, we can extend the UI element with part of the module hash

pexe/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+    let module = sdk
+        .load_module_from_src_actions(script, &names)
+        .map_err(|err| anyhow!("failed to compile plugin: {err}"))?;
+    Ok(format!("{:#}", module.module().batch.id()))
+}
+
+/// Rewrite the `module_hash` line in a manifest's TOML source to the given hash,
+/// preserving formatting of everything else. Adds the line under `[plugin]` if
+/// absent.
+pub fn set_manifest_hash(toml_src: &str, new_hash_hex: &str) -> String {
```
ed255: This looks fragile.  Consider using https://crates.io/crates/toml_edit instead

ed255 (commented):
Overall lgtm!  Take a special look at my first comment, I think we need to address it at some point.

driver/src/pexe_catalog.rs
dhvanipa: yes this is a problem, I will propose a fix in a separate PR

pexe/src/lib.rs
dhvanipa: Agreed, switched to toml_edit

driver/src/pexe_catalog.rs
dhvanipa: tracking here: https://github.com/dobjlabs/zk-craft/issues/80

PR #82, opened 2026-04-22
Support dependent actions

ed255:
Add support of actions that depend on other actions.  In the code and SDK interface I call the action dependency "subaction" which is the same terminology used by the new txlib that Rob has designed.

The Rhai syntax for dependencies looks like this:

```rust
fn Foo(action) {
  var pick = action.subaction("UseWoodPick");
  // ... 
}
```

For now the code assumes a subaction only outputs a single object.

The implementation idea is that an action can execute another action.  This means that we're executing rhai, we jump to a host function, and then we must jump back to executing another rhai function.
To moved the code that executes a rhai function corresponding to an action to the `ActionHandler` and also made the Rhai engine and AST available in the ExeContext, so that we can easly execute rhai code again from a host function defined via an ActionHandler method.

The ActionHandler contains an shared mutable ActionContext.  We need to make sure that when executing a subaction the ActionContext is not borrowed, otherwise we would encounter runtime panics.  You can see this dance here
https://github.com/dobjlabs/zk-craft/blob/cde0f5e5e0f2ddb0eaaf16b500fe5e6aa62896f4/sdk/src/lib.rs#L520-L525
and here
https://github.com/dobjlabs/zk-craft/blob/cde0f5e5e0f2ddb0eaaf16b500fe5e6aa62896f4/sdk/src/lib.rs#L401-L417

---

From a conversation with Rob:

The implementation in the PR assumes subactions go first before anything else.
And there's no safeguard for that, so if you do subactions after a non-subaction operation it will fail in a non-obvious way.
For simplicity we could assume that we want subactions to always go first, and add safeguards for nice error reporting when that's not the case.
In the future, if we find that limiting we could implement support for subactions at any position.

### DIFF
--- a/sdk/README.md
+++ b/sdk/README.md
@@ -147,7 +147,7 @@ The emitted podlang embeds `target` as a hex `Raw(0x00…)` literal.
 - [ ] Execution time type checking without panics
 - [ ] operator+
 - [ ] operator*
-- [ ] dependent action
+- [x] dependent action
 - [x] pexe.zip support (packaged by the `pexe` crate's CLI)
 - [x] manifest support
 - [ ] error pretty print
@@ -211,6 +211,12 @@ UseWoodPick(wood_pick, tx, tx0, private: wood_pick0, wood_pick1, wood_pick2, dur
   tx::TxMutated(tx, tx0, wood_pick, wood_pick0)
 )
 
+MineStoneWithWoodPick(stone, tx, tx0, private: tx1, pick) = AND(
+  UseWoodPick(pick, tx1, tx0)
+  DictContains(stone, "blueprint", "Stone")
+  tx::TxInserted(tx, tx1, stone)
+)
+
 // Classes
 
 IsLog(state, private: tx, tx0) = OR(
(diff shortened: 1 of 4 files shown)


sdk/src/lib.rs
```diff
@@ -390,6 +397,85 @@ impl ActionHandle {
             Ok(())
         }
     }
+    // Execute an action (assumes Execution phase), returns the Action statement
+    fn exe_action(&self, action: &str) -> RuntimeResult<Statement> {
```
ed255, line 401: Most of the code of this function was taken from `Executor::action`.

sdk/README.md
```diff
@@ -230,4 +236,8 @@ IsWoodPick(state, private: tx, tx0) = OR(
   CraftWoodPick(state, tx, tx0)
   UseWoodPick(state, tx, tx0)
 )
+
+IsStone(state, private: tx, tx0) = OR(
```
dhvanipa, line 240: I think we should add support for `MineStoneWithWoodPick` and `MineStoneWithStonePick` to the [rhai script](https://github.com/dobjlabs/zk-craft/blob/main/plugins/craft-basics/plugin.rhai) for completion so we can test end to end (I know you already added a test)

(not blocking merging the PR though, can be in followup)

robknight (approved):

PR #88, opened 2026-04-27
Integrate txlib with new SDK

### LINKED ISSUE #9: Tx layer should verify object type predicate
The Tx layer should verify that each object declares its type (e.g. as a field within the object) and that the type predicate for the declared type is satisfied.

robknight:
Integrates the new txlib as developed in `10object-exploration` with the SDK.

I've removed the higher-level API from txlib in favour of a lower-level API, which allows the SDK to use the `TXBuilder` to accumulate changes as it processes the Rhai scripts, rather than collecting the results and then building the transaction.


driver/Cargo.toml
```diff
@@ -18,7 +18,6 @@ sdk = { path = "../sdk" }
 pexe = { path = "../pexe" }
 pod2utils = { path = "../pod2utils" }
 txlib = { path = "../txlib" }
-timelib = { path = "../timelib" }
```
dhvanipa, line 21: imo we should just delete timelib from main and move it to a branch

Cargo.toml
```diff
@@ -19,7 +18,7 @@ resolver = "2"
 [workspace.dependencies]
 # Branch `next/2026-03-30.a`: Includes:
```
dhvanipa, line 19: can this comment be updated to reflect main

sdk/src/fmt_podlang.rs
```diff
@@ -27,32 +27,72 @@ impl fmt::Display for Intro {
...
+        name.to_string()
+    } else {
+        format!("{name}{ts}")
+    }
+}
+
+/// Two variants encode the protocol/implementation split between
+/// `fmt_chain_at` and `fmt_var_at` so dispatch is at the type level
+/// instead of by string comparison.
+enum VarNameFmt<'a> {
```
ed255: So now we have:
- Chain case: `chain_start, chain_1, chain_2, ..., chain_end`
- Var case: `var0, var1, var2, ..., var`

Is there a reason to have 2 variants, instead of using a generic one for everything, where chain is just a Var case with `name=chain`?
I don't have a very strong opinion on the formatting, `var_start, var_1, var_2, ..., var_end` would be OK to me.

sdk/src/fmt_podlang.rs
```diff
@@ -93,32 +132,28 @@ fn fmt_action_vars(action: &ActionContext) -> Vec<String> {
...
     let pub_var_names = fmt_action_pub_vars(action);
-    for (i, var) in pub_var_names.iter().enumerate() {
-        if i != 0 {
-            write!(w, ", ")?;
-        }
-        write!(w, "{var}")?;
+    for var in &pub_var_names {
+        write!(w, "{var}, ")?;
     }
+    write!(w, "private: ")?;
```
ed255: now there's always at least 2 private args, chain_start and chain_end.  Is this correct?

sdk/src/fmt_podlang.rs
```diff
@@ -135,41 +170,29 @@ fn fmt_action(action: &ActionContext, w: &mut dyn fmt::Write) -> fmt::Result {
...
-                    }
-                    ObjectIO::Output => {}
-                    ObjectIO::Mutate => {
-                        writeln!(w, "  Is{}({})", class, ArgFmt(&vars, obj))?;
-                    }
-                }
+            Inst::Object { io, obj, .. } => {
+                // Input/Mutate type-guards are enforced at replay time by
+                // the new txlib (ReplayDelete/ReplayMutate call the obj's
+                // type predicate). Nothing to emit inline.
```
ed255: But the action needs to say what kind of object it expects to consume, something like:
```
Contains(obj, "type", IsClass)
```
where is this done?

sdk/src/tests.rs
```diff
@@ -110,57 +110,46 @@ fn test_sdk_1() {
         .load_module_from_src_actions(craft_src, actions)
         .unwrap();
 
-    fn list(xs: &[(&str, &str)]) -> Vec<(String, String)> {
-        xs.into_iter()
-            .map(|(a, b)| (a.to_string(), b.to_string()))
-            .collect()
+    fn classes<'a>(refs: impl Iterator<Item = &'a ActionObjectRef>) -> Vec<&'a str> {
```
ed255, line 105: thanks for renaming this, I was very uninspired when I called this `list`

sdk/src/tests.rs
```diff
@@ -110,57 +110,46 @@ fn test_sdk_1() {
         .load_module_from_src_actions(craft_src, actions)
```
ed255, line 102: Could you update the `sdk/README.md` with the new example podlang code generated from `test_sdk_1`?
This is the content:
```
use module 0x3a76c2f1f39cd4e3e58c2f52f09e9bf6d28310197dc479a4a6f176eb2271ab7d as tx
use intro Vdf(count, input, output) from 0xb77a964de74c8569e6c6172692bb50147df9334fd9b572abc8d4d9c688a40e06
use intro LtEqU256(lhs, rhs) from 0x2e79114ee823f4783ab5b6eb93b49abba87fb69b4d14de4cf1d78648ade73529

// Actions

FindLog(log, chain_start, chain_end, private: log0, work) = AND(
  DictContains(log0, "blueprint", "Log")
  Vdf(3, log0, work)
  DictUpdate(log, log0, "work", work)
  tx::TxInserted(chain_end, chain_start, log)
)

CraftWood(log, wood, chain_start, chain_end, private: chain_1, wood0, key) = AND(
  DictContains(wood0, "blueprint", "Wood")
  DictUpdate(wood, wood0, "key", key)
  LtEqU256(wood, Raw(0x0020000000000000000000000000000000000000000000000000000000000000))
  tx::TxDeleted(chain_1, chain_start, log)
  tx::TxInserted(chain_end, chain_1, wood)
)

CraftSticks(wood, stick_a, stick_b, chain_start, chain_end, private: chain_1, chain_2) = AND(
  DictContains(stick_a, "blueprint", "Stick")
  DictContains(stick_b, "blueprint", "Stick")
  tx::TxDeleted(chain_1, chain_start, wood)
  tx::TxInserted(chain_2, chain_1, stick_a)
  tx::TxInserted(chain_end, chain_2, stick_b)
)

CraftWoodPick(wood, stick, pick, chain_start, chain_end, private: chain_1, chain_2) = AND(
  DictContains(pick, "blueprint", "WoodPick")
  DictContains(pick, "durability", 100)
  tx::TxDeleted(chain_1, chain_start, wood)
  tx::TxDeleted(chain_2, chain_1, stick)
  tx::TxInserted(chain_end, chain_2, pick)
)

UseWoodPick(wood_pick, chain_start, chain_end, private: wood_pick0, wood_pick1, wood_pick2, durability, key, work) = AND(
  Gt(wood_pick0.durability, 0)
  SumOf(wood_pick0.durability, durability, 1)
  DictUpdate(wood_pick1, wood_pick0, "durability", durability)
  DictUpdate(wood_pick2, wood_pick1, "key", key)
  Vdf(10, wood_pick2, work)
  DictUpdate(wood_pick, wood_pick2, "work", work)
  tx::TxMutated(chain_end, chain_start, wood_pick, wood_pick0)
)

MineStoneWithWoodPick(stone, chain_start, chain_end, private: chain_1, pick) = AND(
  UseWoodPick(pick, chain_start, chain_1)
  DictContains(stone, "blueprint", "Stone")
  tx::TxInserted(chain_end, chain_1, stone)
)

// Classes

IsLog(state, chain_start, chain_end, private: _other_0) = OR(
  FindLog(state, chain_start, chain_end)
  CraftWood(state, _other_0, chain_start, chain_end)
)

IsWood(state, chain_start, chain_end, private: _other_0, _other_1) = OR(
  CraftWood(_other_0, state, chain_start, chain_end)
  CraftSticks(state, _other_0, _other_1, chain_start, chain_end)
  CraftWoodPick(state, _other_0, _other_1, chain_start, chain_end)
)

IsStick(state, chain_start, chain_end, private: _other_0, _other_1) = OR(
  CraftSticks(_other_0, state, _other_1, chain_start, chain_end)
  CraftSticks(_other_0, _other_1, state, chain_start, chain_end)
  CraftWoodPick(_other_0, state, _other_1, chain_start, chain_end)
)

IsWoodPick(state, chain_start, chain_end, private: _other_0, _other_1) = OR(
  CraftWoodPick(_other_0, _other_1, state, chain_start, chain_end)
  UseWoodPick(state, chain_start, chain_end)
)

IsStone(state, chain_start, chain_end) = OR(
  MineStoneWithWoodPick(state, chain_start, chain_end)
)
```

sdk/src/lib.rs
```diff
@@ -300,26 +328,44 @@ struct VarState {
...
+    /// User-defined vars (objects, intro outputs, temporaries). The
+    /// txlib chain var is tracked separately in `chain_ts` because its
+    /// rendering follows a different (protocol-defined) naming scheme.
     vars: Vec<String>,
     var_state: HashMap<String, VarState>,
-    exe_ctx: Option<ExeContext>,
+    /// Timestamp of the txlib chain var. Bumped once per direct event
+    /// (insert/delete/mutate) and once per sub-action call.
+    chain_ts: usize,
+    exe_ctx: Option<Rc<RefCell<ExeContext>>>,
```
ed255, line 344: from this change I infer that you needed to take a reference copy of the ExeContext from an ActionContext.

sdk/src/lib.rs
```diff
@@ -344,47 +393,62 @@ impl ActionContext {
...
                     arg.borrow_mut().set_value(Value::from(obj));
                 }
                 ObjectIO::Mutate => {
-                    let (st_class, obj) = exe_ctx.inputs[exe_ctx.input_index].clone();
-                    exe_ctx.input_index += 1;
-                    exe_ctx.sts.push(st_class);
-                    arg.borrow_mut().set_value(Value::from(obj));
+                    let obj = exe_ctx.inputs.pop().expect("exists");
+                    arg.borrow_mut().set_value(Value::from(obj.clone()));
+                    original = Some(obj);
```
ed255, line 430: nice, instead of using the `exe_ctx.inputs` twice, now you just pop from it and make a copy for mutate cases.

ed255: Previously an input object would come with a pod that contained the IsClass statement.
I think we need to figure out an equivalent.  Probably we want to use a statement like
```
Contains(obj, "type", @self_predicate(IsClass))
```
If we have the contents of the object we can generate this statement on the go.

sdk/src/lib.rs
```diff
@@ -483,14 +603,17 @@ impl ActionHandle {
...
                 .bld
                 .builder
-                .priv_op(Operation(op_type.clone(), args, OperationAux::None))
-                .unwrap();
-            exe_ctx.sts.push(st);
+                .priv_op(Operation(op_type.clone(), op_args, OperationAux::None))
+                .unwrap()
+        });
+        if let Some(st) = st {
+            ctx.sts.push(st);
```
ed255: I think we should move `sts` to `ExeContext`.  It will make the code clearer.

sdk/src/lib.rs
```diff
@@ -507,36 +630,68 @@ impl ActionHandle {
...
-        // variables.  We could define the var syntax with destructuring of arrays with
-        // `engine.register_custom_syntax_with_state_raw`
-        let arg = Rc::new(RefCell::new(VarOrValue::var(Type::Dict)));
-        // We don't borrow the ActionContext here because exe_action requires exclusive access to
-        // it so that it can execute the Rhai code that calls hosts functions which borrow and
-        // borrow_mut the ActionContext.
-        let is_exe_phase = self.0.borrow().exe_ctx.is_some();
-        let st_action = if is_exe_phase {
-            self.exe_action(&action)?
+        let exe_rc_opt = self.0.borrow().exe_ctx.clone();
```
ed255, line 671: This is an example of getting a reference copy of the ExeContext without having an open reference to the ActionContext.
This looks cleaner than before, and justifies having Rc<RefCell<ExeContext>> inside ActionContext

sdk/src/lib.rs
```diff
@@ -507,36 +630,68 @@ impl ActionHandle {
...
+            let sub_handle = ActionHandle::new(action.clone(), Some(exe_rc.clone()));
+            let st_sub = sub_handle.exe_action()?;
+
+            // Reveal so per-output IsX pods can reference the
+            // sub-action's Action statement via the internal pod.
+            exe_rc.borrow_mut().bld.builder.reveal(&st_sub).unwrap();
+
+            // Drop the sub handle's shared exe_ctx so it doesn't pin
+            // the shared ExeContext past this call.
+            sub_handle.0.borrow_mut().exe_ctx = None;
```
ed255: Is this really needed?  `sub_handle` is dropped at the end of the code block already.

sdk/src/lib.rs
```diff
@@ -623,22 +775,20 @@ impl ActionHandle {
...
+                LtEqU256Pod::new_boxed(&exe_ctx.params, exe_ctx.vd_set.clone(), l, r)
             }
             .unwrap();
-            let st_lt_eq_u256 = lt_eq_u256_pod.pub_statements()[0].clone();
-            exe_ctx
-                .bld
-                .builder
-                .add_pod(exe_ctx.main_pod(lt_eq_u256_pod))
-                .unwrap();
-            exe_ctx.sts.push(st_lt_eq_u256);
```
ed255, line 641: oh, `sts` was in `ExeContext` before.  Why is it in `ActionContext` now?

sdk/src/lib.rs
```diff
@@ -871,68 +1041,90 @@ fn try_value_from_dynamic(v: Dynamic) -> RuntimeResult<Value> {
...
+/// One object reference in an action, in declaration order. Only
+/// `class` is exposed; `io` is internal — used by the
+/// `inputs()`/`outputs()` filters.
+#[derive(Debug, Clone)]
+pub struct ActionObjectRef {
+    io: ObjectIO,
+    pub class: String,
+}
+
+/// Collected metadata that declares an Action.
```
ed255, line 1066: note to myself: renaming of:
- local -> direct
- total -> aggregated

sdk/src/lib.rs
```diff
@@ -871,68 +1041,90 @@ fn try_value_from_dynamic(v: Dynamic) -> RuntimeResult<Value> {
...
+    pub fn total_inputs(&self) -> impl Iterator<Item = &ActionObjectRef> {
+        self.aggregated_inputs.iter()
+    }
+
+    /// Object refs produced by this action plus any transitively-called
+    /// sub-actions, in declaration order. Used for action-signature
+    /// reporting and output-slot validation by the driver.
+    pub fn total_outputs(&self) -> impl Iterator<Item = &ActionObjectRef> {
+        self.aggregated_outputs.iter()
+    }
```
ed255: It would be clearer if we're consistent with naming:
- I'd rename `inputs` and `outputs` to `prefix_inputs` and `prefix_outputs` to remove ambiguitiy
- Either use `direct/aggregated` or `local/total` but not a mix of both.

sdk/src/lib.rs
```diff
@@ -1242,103 +1466,142 @@ impl Executor {
...
     /// Execute an action that consumes some input objects and produces some output objects
     pub fn action(
         &self,
         action: &str,
         inputs: Vec<SpendableObject>,
     ) -> Result<SpendableObjects, SdkError> {
         // TODO: In this function: return errors instead of panic from unwrap.
-        let action = self.module.action_by_name(action);
+        let action_name = self.module.action_by_name(action).name.clone();
```
ed255: oh, this is very redundant, this line could be removed

sdk/src/lib.rs
```diff
@@ -1242,103 +1466,142 @@ impl Executor {
...
+
+        // Release the handle's Rc clone so `exe_rc` has a unique
+        // owner for the `try_unwrap` below.
+        action_handle.0.borrow_mut().exe_ctx = None;
+        let ExeContext {
+            tx_builder,
+            mut bld,
+            outputs,
+            ..
+        } = Rc::try_unwrap(exe_rc)
```
ed255, line 1564: I considered this in my initial implementation but I was afraid of having some copy of the `Rc` somewhere.  I'm glad this works, because it's more straightforward!

ed255 (commented):
First pass review

sdk/src/fmt_podlang.rs
robknight: Only that chain is constant, and so this avoids having to use a special string in the code. Maybe it would be simpler to just use the string.

sdk/src/fmt_podlang.rs
robknight: They are public, but yes.

ed255: oh ok, then I got confused.
The previous code was able to handle the case when there are no private args, but the current one doesn't.
This was achieved with this part:
```rust
    if var_names.len() - pub_var_names.len() > 0 {
        write!(w, ", private: ")?;
    }
```
and also doing iter enum on `pub_var_names` to add commas everywhere except for the last entry

sdk/src/lib.rs
robknight: So my thinking was to make `ActionContext` manage things that happen with the action's frame, to avoid the situation where sub-actions need to occur before any other mutation/insertion/deletion. However, although that justifies making `ActionContext` responsible for managing its own `insts`, we could allow `sts` to remain on `ExeContext`. I'm not strongly opinionated - my instinct was to break things up into structurally separate units rather than to have each action manipulate shared data, but it's not really *necessary* to do it that way. I can move it back.

sdk/src/lib.rs
robknight: I've switched this to `local` and `total`.

sdk/src/lib.rs
robknight: Removed

sdk/src/lib.rs
robknight: You're right, removed.

sdk/src/lib.rs
ed255: I see.
One point of view I had was that `ExeContext` is where things that only happen at Execution time are stored.
And the generation of statements would land there for this reason.
But I understand your point of view that an ActionContext seems to be local to a particular action whereas the ExeContext lasts over actions and nested subactions.  But on the other hand, the current implementation passes a clone of the ActionContex to the ActionHanlder used in a subaction.

So I'm more inclined to say that it makes more sense to me for `sts` to remain in `ExeContext`.  This way there's no need to have these blocks:
```
...
```
I was looking for an example and saw that you've already moved `sts` to `ExeContext`.

sdk/src/fmt_podlang.rs
robknight: I've fixed this by restoring something pretty close to the original code.

txlib/src/predicates/crafting_test.podlang
dhvanipa, line 1: Do we need this to be merged in on main? Or can we remove it?

sdk/src/tests.rs
```diff
@@ -40,15 +40,13 @@ fn test_sdk_1() {
     let craft_src = r#"
         fn FindLog(action) {
             var log = action.output("Log");
-            log.set([["blueprint", "Log"]]);
```
dhvanipa, line 43: nice I like how blueprint was removed here! Can it also be updated in `plugins/craft-basics/plugin.rhai` which is the end to end test?

txlib/src/predicates/crafting_test.podlang
dhvanipa: I guess we have 3 different things here:
1) sdk/src/tests with a bunch of crafting podlang to test the SDK
2) plugins/craft-basics/plugin.rhai which is the crafting pexe used to test the end to end flow
3) (new) txlib/src/predicates/crafting_test.podlang which is a bunch of crafting podlang to test the txlib

tbh I can see arguments for keeping all 3, so up to you

txlib/src/predicates/txlib.podlang
```diff
@@ -1,114 +1,346 @@
...
+// values for those two fields (they pass through ReplayActions
+// verbatim into `tx_final`, since ReplayAction only updates the
+// `live` and `nullifiers` keys).
+TxFinalized(state_root_hash, tx_final, nullifiers,
+    private: before_ctx, chain_start, chain_final) = AND(
+  InputsGrounded(before_ctx.live, state_root_hash)
+  HashOf(chain_start, before_ctx.live, 0)
+  DictInsert(before_ctx, {"nullifiers": {}, "tx_start": 0, "tx_end": 0}, "live", before_ctx.live)
+  DictContains(tx_final, "nullifiers", nullifiers)
+  ReplayActions(before_ctx, tx_final, chain_start, chain_final)
```
ed255: These predicates would be clearer with a consistent usage of `tx`, `ctx` and `chain`.
`ReplayActions` takes a before and after ctx, but here the names are `before_ctx` and `tx_final`.

Similarly inside context there are two fields named `tx_start` and `tx_end` but they get assigned wildcards named `ctx`

ed255 (approved):
LGTM! I've just left one more comment regarding naming in `txlib.podlang`

PR #89, opened 2026-04-27
add stone and stone actions to craft basics pexe 

dhvanipa:
We switched to pexe but didnt port over the stone and stonepick example we had before. We should keep it to be able to test end to end

Fixed some frontend types missed in https://github.com/dobjlabs/zk-craft/pull/91


dhvanipa:
Merging in to fix a bug on main, but will still address any comments from your review @ed255

ed255 (commented):
LGTM!

PR #90, opened 2026-04-28
fix subaction outputs

dhvanipa:
The engine accumulates outputs from subactions and the parent into the same `exe_ctx.outputs`, but `ActionMeta::from_action_ctx` only copied subaction inputs into the parent's metadata,. It never outputs, so the descriptor undercounted by one per subaction.

So `MineStoneWithWoodPick` failed in the GUI with `output mismatch: descriptor expects 1, engine returned 2`

### DIFF
--- a/sdk/src/fmt_podlang.rs
+++ b/sdk/src/fmt_podlang.rs
@@ -104,9 +104,7 @@ fn fmt_action_pub_vars(action: &ActionContext) -> Vec<String> {
             // ActionMeta accumulates it. The class predicate (`IsClass`) calls
             // this predicate with one positional arg per output, so the parent
             // must publish the subaction's binding alongside its own outputs.
-            Inst::SubAction { action: _, obj } => {
-                vars.push(obj.borrow().var_name().to_string())
-            }
+            Inst::SubAction { action: _, obj } => vars.push(obj.borrow().var_name().to_string()),
             _ => {}
         }
     }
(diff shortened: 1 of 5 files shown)


sdk/src/fmt_podlang.rs
```diff
@@ -93,13 +93,19 @@ fn fmt_action_vars(action: &ActionContext) -> Vec<String> {
...
-        } = inst
-        {
-            vars.push(obj.borrow().var_name().to_string())
+        match inst {
+            Inst::Object {
+                io: ObjectIO::Mutate | ObjectIO::Output,
+                obj,
+                class: _,
+            } => vars.push(obj.borrow().var_name().to_string()),
+            // A subaction's primary output is an output of the parent action
```
ed255, line 102: This was not the intended design.
The idea was that an action only has the explicit outputs as public arguments.
The outputs from subactions don't need to be public arguments, because the classes of the subaction outputs just use the subaction statements.

So I think the confusion appears from naming.
- If I say `action.outputs`, do I mean just the explicit outputs of the action?  Or do I include the nested subactions outputs too?  We may need a different name for each case.

ed255: The same happens with inputs BTW.

ed255: Ok here's my proposal to fix this issue, using "local/total" naming to distinguish each case https://github.com/dobjlabs/zk-craft/pull/91

PR #91, opened 2026-04-29
distinguish local vs total inputs/outputs in actions

ed255:
The meaningful changes are in `sdk/src/tests.rs` where I extend the test to verify the generated `ActionMeta`s and `sdk/src/lib.rs` where I introduce 2 new fields in `ActionMeta`.

Then I renamed 3 fields of `ActionSummary` in `driver`, `Action` in `mcp` and `Action` in the `gui` for clarity, so all the other files that change in this PR are just from this renaming.


dhvanipa (approved):
lgtm!

Issue #80, opened 2026-04-21
catalog conflates same-named classes/actions across plugins


dhvanipa:
`PexeCatalog::from_plugins` in `driver/src/pexe_catalog.rs` keys its internal tables by bare `name: String`. When two `.pexe` plugins declare a class or action with the same name, the catalog collapses them, silently discarding metadata from one, and, more seriously, cross-wiring action-to-class hash links across module boundaries.

Within a single pod2 module, class/action names are unique. Across modules, two classes named `Wood` are different: different `module_hash` → different `class_hash` → the verifier treats them as distinct predicates. The catalog should do the same.

Originally mentioned: https://github.com/dobjlabs/zk-craft/pull/79#discussion_r3116367653

PR #92, opened 2026-05-01
add docs

dhvanipa:
https://dobjlabs.github.io/zk-craft/

docs outlining the digital objects system (as it's designed & implemented, no future stuff)

Stuff not added:
- relayer (not sure how important it is to explain it, I think its simpler to just blackbox it as posting to the chain at this stage. If it did something like groth16 compression then I think it'd be important to include)
- MCP (imo, we should add this when we have a more clear bot story), docs already mention driver could be called from mcp

Upcoming changes that will change these docs:
- identity/object id
- identify an action by module + action id, vs just action id, since there can be multiple (https://github.com/dobjlabs/zk-craft/issues/80)

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,57 @@
+name: Deploy docs to GitHub Pages
+
+on:
+  push:
+    branches: ["**"]
+    paths:
+      - "docs/**"
+      - ".github/workflows/docs.yml"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+concurrency:
+  group: pages
+  cancel-in-progress: false
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: docs
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: pnpm/action-setup@v4
+        with:
+          version: 9
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          cache: pnpm
+          cache-dependency-path: docs/pnpm-lock.yaml
+
+      - run: pnpm install --frozen-lockfile
+
+      - run: pnpm build
+
+      - uses: actions/configure-pages@v5
+
+      - uses: actions/upload-pages-artifact@v3
+        with:
+          path: docs/dist
+
+  deploy:
+    needs: build
+    runs-on: ubuntu-latest
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    steps:
+      - id: deployment
+        uses: actions/deploy-pages@v4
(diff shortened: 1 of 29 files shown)


.github/workflows/docs.yml
```diff
@@ -0,0 +1,57 @@
+name: Deploy docs to GitHub Pages
+
+on:
+  push:
+    branches: ["**"]
```
dhvanipa, line 5: will change this to main before merging, its like this for easier testing right now

docs/pages/actions/transactions.mdx
```diff
@@ -0,0 +1,110 @@
...
+| **Action scopes**         | A way to group events by which action emitted them. Scopes can nest — a sub-action's events appear nested under its parent. The whole tree is proven as one transaction.                                       | The replay walks the tree; this is what makes "sub-action" different from "called function in the same scope." |
+| **`live` / `nullifiers`** | The two sets that summarize the transaction's effect. `live` is every object that exists at the end (carryovers + new outputs); `nullifiers` is the hash of every consumed input.                              | Anyone verifying the transaction; the synchronizer indexes nullifiers to prevent double-spends.                |
+| **`tx_final`**            | The hash of `ctx` (the dictionary bundling `live` and `nullifiers`). The single value that travels with the proof to the chain.                                                                                | The synchronizer indexes it; later transactions reference it via grounding witnesses.                          |
```
dhvanipa, line 55: Maybe we dont want to call this "tx" as it conflicts with the notion of transactions on Ethereum, eg there is both "tx_hash" and "tx_final" which mean 2 different things here

dhvanipa, line 55: in the current design, we post the tx hash, but another option is to post the object hash itself, and the synchronizer keeps a set of object hashes

This would be a bigger change, but worth discussing as it could simplify things conceptually, although introduces a future scaling issue

ed255: I guess you mean posting the objects hashes for the created objects in the action right? (that includes inserted and the output of mutations).
We actually talked about this with Rob last week.  We think the whole design of the system would be simpler and more transparent this way.  Then the synchronizer keeps track of an explicit set of created objects (similar to how it currently tracks an explicit set of consumed objects via the nullifier set)

robknight: "Transaction" is the correct CS term, and the overload is annoying (in the early design ideation, we explicitly avoided thinking about Ethereum as the consensus layer, so the overload wasn't relevant). My other possible name was "[Unit of Work](https://martinfowler.com/eaaCatalog/unitOfWork.html)", but this is a bit verbose. "Batch" has the wrong connotation because most batch-processing systems do not have atomicity (it's typically possible for three out of five batched writes to succeed, for example). Perhaps "commit" as both a noun and a verb could work, in the same way that git uses it. I still think the database transaction analogy is better, though.

dhvanipa: I'm curious what the potential downsides of it are

PR #94, opened 2026-05-04
distinguish same-named classes/actions across plugins

### LINKED ISSUE #80: catalog conflates same-named classes/actions across plugins

dhvanipa:
Fixes https://github.com/dobjlabs/zk-craft/issues/80

- Qualified id `<plugin>::<class|action>` (e.g. `craft-basics::CraftWood`) is the canonical handle on disk, in MCP, and across the Tauri wire. Plugin-name uniqueness is enforced at catalog load.
- Display name is the bare label, paired with `pluginName`
- The cryptographic source of truth remains `obj["type"]`, ie the on-chain `Is{class}` predicate hash, which is module-scoped because it derives from the plugin's `module_hash`.

### DIFF
--- a/driver/src/execute.rs
+++ b/driver/src/execute.rs
@@ -112,7 +112,7 @@ pub(crate) fn validate_execute_request(
     if input.input_objects.len() != action.total_inputs.len() {
         return Err(anyhow!(
             "{} expects {} inputs, got {}",
-            input.action_id,
+            input.action,
             action.total_inputs.len(),
             input.input_objects.len()
         ));
@@ -157,16 +157,16 @@ pub(crate) fn resolve_inputs(
                 entry.record.id
             ));
         }
-        // Belt-and-suspenders: compare both the qualified class id stored on
+        // Belt-and-suspenders: compare both the qualified class stored on
         // disk AND the on-chain `obj["type"]` predicate hash. Mismatch on
-        // either is fatal — the second check catches files whose class_id
+        // either is fatal — the second check catches files whose class
         // text drifted from the actual pod-level identity.
-        if entry.record.class_id != required.id {
+        if entry.record.class != required.class {
             return Err(anyhow!(
                 "input class mismatch for {}: expected {}, got {}",
                 entry.record.id,
-                required.id,
-                entry.record.class_id
+                required.class,
+                entry.record.class
             ));
         }
         let actual_class_hash = obj_type_hash(&entry.record.obj).ok_or_else(|| {
@@ -196,27 +196,6 @@ pub(crate) fn obj_type_hash(obj: &pod2::middleware::containers::Dictionary) -> O
     Some(Hash(value.raw().0))
 }
 
-/// Build the lowercase filename prefix for a `.dobj` of the given qualified
-/// class id (`<plugin>::<class>`). Plugin names are already restricted to
-/// `[A-Za-z0-9_-]` at catalog load, but class names come from arbitrary
-/// rhai string literals (e.g. `action.output("…")`) so they could in
-/// principle contain path-significant characters. To keep written files
-/// inside `~/.dobj/objects/`, every char outside the allowlist
-/// `[a-z0-9_-]` (after lowercasing) is replaced with `_`.
-pub(crate) fn file_prefix_for_class(class_id: &str) -> String {
-    class_id
-        .chars()
-        .map(|c| {
-            let lower = c.to_ascii_lowercase();
-            if lower.is_ascii_alphanumeric() || lower == '-' || lower == '_' {
-                lower
-            } else {
-                '_'
-            }
-        })
-        .collect()
-}
-
 #[derive(Debug)]
 pub(crate) struct SavedFiles {
     pub(crate) output_files: Vec<String>,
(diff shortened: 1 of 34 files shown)


ed255:
In podlang we use `::` for qualified names.  Perhaps doing the same here is a good idea to have consistency (so that we don't have to wonder when to use `::` and when to use `:`)

driver/src/execute.rs
```diff
@@ -171,6 +191,32 @@ pub(crate) fn resolve_inputs(
...
+    Some(Hash(value.raw().0))
+}
+
+/// Build the lowercase filename prefix for a `.dobj` of the given qualified
+/// class id (`<plugin>:<class>`). Plugin names are already restricted to
+/// `[A-Za-z0-9_-]` at catalog load, but class names come from arbitrary
+/// rhai string literals (e.g. `action.output("…")`) so they could in
+/// principle contain path-significant characters. To keep written files
+/// inside `~/.dobj/objects/`, every char outside the allowlist
+/// `[a-z0-9_-]` (after lowercasing) is replaced with `_`.
```
ed255: This is a good catch, we don't want path traversal exploits!  I think we should be more strict with this at the SDK level, where the SDK would reject class names that have characters outside of `[a-z0-9_-]`

driver/src/pexe_catalog.rs
```diff
@@ -5,41 +5,57 @@
...
+    classes_by_id: HashMap<String, CatalogClass>,
+    classes_by_hash: HashMap<Hash, String>,
     combined_podlang_src: String,
     mock_proofs: bool,
 }
 
+pub fn qualified_id(plugin_name: &str, name: &str) -> String {
+    format!("{plugin_name}:{name}")
+}
+
```
ed255: I noticed that many times in this PR you have 3 strings:
- plugin_name
- name
- id

whenever you we have these 3 strings we assume the following invariant holds: `id = {plugin_name}:{name}`, but it's not straight forward to reason that this invariant indeed holds (you need to review all the code carefully).
Perhaps there's no need to have an explicit `id` beyond the presentation layer.  I would suggest having a struct like `QualifiedClass { plugin_name: String, name: String}` and using it instead of `id`.
So instead of comparing by id you compare by `QualifiedClass`.  Instead of indexing by id you index by `QualifiedClass`.
APIs that query classes receive `QualifiedClass`.
And then only the frontend renders `plugin_name:name` when showing classes to the user, and parses `plugin_name:name` into `QualifiedClass{plugin_name, name}` when reading classes from user input. 

The benefits of this approach is:
- You remove one redundant field in many structs `id`
- You don't need to reason about the invariant holding because `id` is the output of a deterministic function, not a third field
- Richer types in the driver: For example this
```
pub struct ObjectSummary {
    pub id: String,
    pub file_name: String,
    /// Qualified class id (`<plugin>:<class>`).
    pub class_id: String,
    /// Bare class name from the producing plugin's manifest.
    pub class_display_name: String,
    pub plugin_name: String,
    pub class_hash: String,
    pub status: ObjectStatus,
    pub tx_hash: Option<String>,
    pub grounded: Option<bool>,
    pub fields: HashMap<String, serde_json::Value>,
}
```
can just be
```
pub struct ObjectSummary {
    pub id: String,
    pub file_name: String,
    pub class: QualifiedClass,
    pub class_hash: String,
    pub status: ObjectStatus,
    pub tx_hash: Option<String>,
    pub grounded: Option<bool>,
    pub fields: HashMap<String, serde_json::Value>,
}
```
  - No need to add comments because the type already describes what the `class` field is
  - You can do `summary.class.id()` or `summary.class.display_name()`, etc.
- You can add convenient methods to `QualifiedClass` like `id()`

ed255 (commented):
Overall looks good to me but I'd like to strongly suggest using `plugin::name` instead of `plugin:name` to be consistent with podlang (and avoid the mental overhead of dealing with 2 different separators in the project), and also add a structure to carry qualified names instead of having the id explicitly stored in strings, at least in the driver side.

driver/src/execute.rs
dhvanipa: added!

driver/src/pexe_catalog.rs
dhvanipa: updated!

dhvanipa:
> Overall looks good to me but I'd like to strongly suggest using `plugin::name` instead of `plugin:name` to be consistent with podlang (and avoid the mental overhead of dealing with 2 different separators in the project), and also add a structure to carry qualified names instead of having the id explicitly stored in strings, at least in the driver side.

Yup, updated both.

ed255 (approved):
LGTM!

Issue #63, opened 2026-03-30
simplify synchronizer & driver API to use RawValue/Hash instead of String


dhvanipa:
See https://github.com/dobjlabs/zk-craft/pull/60#discussion_r3008836981

also see https://github.com/dobjlabs/zk-craft/pull/95#discussion_r3267054006

PR #95, opened 2026-05-06
decouple driver into a standalone daemon with CLI + MCP clients

dhvanipa:
Today the driver runs in-process inside the Tauri desktop app and inside a separate dobjd-equivalent for any other consumer. This PR collapses everything onto one driver process with multiple thin clients — the desktop app, the website, the new terminal CLI, and the MCP server all talk to the same daemon over HTTP/SSE.

        ┌─ desktop (Tauri webview shell) ─┐
        ├─ browser tab ───────────────────┤   HTTP/SSE
        ├─ `dobj` CLI ────────────────────┼──────────►  dobjd  ──►  ~/.dobj/
        └─ MCP client (Claude, etc.) ─────┘            (one process)
        
## What's new

`dobjd` — the long-running daemon. Wraps Arc<Driver> behind:
- HTTP API on `:7717` — REST routes mirroring driver capabilities, plus an SSE event hub on `/events`
- MCP server on `:7718/mcp` — moved from `app-gui/src-tauri/src/mcp.rs`, shares the same broadcast hub so MCP-driven actions show up in real time on every other connected client
- `bitcraft-mcp-proxy` — stdio↔HTTP bridge for Claude Desktop (which only speaks stdio MCP)

I made a separate crate for this because the driver crate is currently a pure library — pod2/plonky2/reqwest only — and folding dobjd in would add axum, tower-http, notify, tokio-broadcast to every consumer's dep graph

I picked HTTP so you could use a website GUI to generate proofs vs having to download a desktop app (you can test with `just dev` to see). Security is not added in this version (but modern browsers do prompt the user to allow websites to connect to localhost vs allowing any to connect). I think in the future we could add a token based auth which is standard in other places 

I think one process is better vs a separate process for each so the shared logic only runs once, and in the future we could for eg prevent two actions from running in parallel that consume the same input as we know one will fail.

`cli` — terminal CLI. 

SKILL.md — agent-driven install guide. Pasteable prompt for any MCP-aware agent:
>Read https://bitcraft.s3.us-east-2.amazonaws.com/v0.1.0-rc.17/SKILL.md and set up bitcraft

This is only the first pass of a SKILL.md to see what it feels like, but it will be iterated upon. The link is to a S3 bucket because this repo is private so you cannot curl releases from it. This will be changed in the future

Updated release action for daemon (removed desktop app for now so its faster to test and the current gui is dev only anyways)

Removed a bunch of stuff that felt like we don't have a clear story for why we need right now. Given that the system is sufficiently complex, removing unnecessary stuff lowers mental overhead of working with the code. For the same reasons, consolidated duplicated types.

All naming is TBD and we will need a follow up PR where we (as a team) agree of the names for everything: driver, pexe, dobj, etc.

## API parity across surfaces
image


driver/src/types.rs
```diff
@@ -20,18 +20,10 @@ pub struct DriverSettings {
     pub relayer_api_url: String,
 }
 
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub enum ObjectSelector {
```
dhvanipa, line 25: for simplicity, removed support for any API using object id. You have to pass in an object path, either relative to the dobj home directory or absolute.

synchronizer/build.rs
dhvanipa, line 1: I had to add these build files because otherwise the release binaries were crashing with not being able to find scip libraries. Synchronize and relayer import pod2 library but dont use the pod builder, however scip is not a feature that can be disabled/enabled.

dhvanipa:
For some reason the MCP doesnt show any logs when a tool is running, but with the events SSE, Claude can build a simple HTML to stream it

eg 
image

dhvanipa:
@robknight @ed255 adding both of you as reviewers to this because this is an important change and at minimum you should be aware of it

app-gui/src/shared/api/tauriClient.ts
```diff
@@ -16,83 +34,185 @@ export type {
...
+  }
+  return (await res.json()) as T;
+}
+
+// === Driver-backed operations: always HTTP ==================================
+
+// Inventory and the action catalog are independent reads; callers run them
+// in parallel via Promise.all rather than letting one block the other.
+export function loadInventory(): Promise<InventoryObjectPayload[]> {
+  return fetch(`${HTTP_BASE}/inventory`).then(
```
robknight: Our `fetch` calls have no AbortController, timeout, or error handling here, so if a fetch fails for any reason then the caller will hang indefinitely.

Since the calls are presumably local, this might be an acceptable tradeoff, but even a crude "oh no, the dobjd has crashed/been killed" error handler might be worthwhile.

app-gui/src-tauri/src/objects/watcher.rs
robknight, line 1: Without the watcher, updates to the object collection rely on the 4s polling timer in dobj, right? I think that's OK but I want to confirm my understanding.

cli/README.md
```diff
@@ -0,0 +1,125 @@
...
+```
+
+Two flavors of commands.
+
+### API wrappers — every driver capability through one HTTP/SSE endpoint
+
+| Command | Hits |
+|---|---|
+| `inventory` | `GET /inventory` |
+| `actions` | `GET /inventory` (filters to actions only) |
```
robknight: I think this hits `GET /actions` not `GET /inventory`

SKILL.md
```diff
@@ -0,0 +1,166 @@
...
+### 2. Create the dobj home
+
+```bash
+mkdir -p ~/.dobj/bin ~/.dobj/actions
+```
+
+### 3. Download `dobjd` and `dobj`
+
+```bash
+RELEASE=https://bitcraft.s3.us-east-2.amazonaws.com/v0.1.0-rc.17
```
robknight, line 47: Is there a risk that users hit this skill file even after a newer release occurs? If so they will continue trying to install an old version.

cli/src/commands.rs
```diff
@@ -0,0 +1,444 @@
...
+            render_inputs(&action.total_inputs),
+            action.description,
+        );
+    }
+    Ok(())
+}
+
+pub async fn state_root(client: &DobjdClient) -> Result<()> {
+    let root = client.get_text("/state-root").await?;
+    println!("{}", root.trim_matches('"'));
```
robknight: `serde_json::from_str::<String>` is a safer way to extract string contents from a JSON string.

dobjd/src/routes/events.rs
```diff
@@ -0,0 +1,39 @@
...
+/// auto-reconnects, and on reconnect the client refetches state via the
+/// regular REST routes.
+pub async fn stream(
+    State(state): State<AppState>,
+) -> Sse<impl Stream<Item = Result<SseEvent, Infallible>>> {
+    let rx = state.events.subscribe();
+    let stream = BroadcastStream::new(rx).filter_map(|item| async move {
+        match item {
+            Ok(event) => {
+                let json = serde_json::to_string(&event).ok()?;
```
robknight: Is there any risk that this might not parse successfully? Probably not if we're sure that the server only ever emits valid JSON.

robknight (approved):
This is a big one and I can't claim to have reviewed it line-by-line.

However, the basic architecture makes sense and the changes seem self-contained, so overall this looks good to me. The comments I've left are not blockers but could either be fixed quickly or left as issues for later follow-up.

cli/src/commands.rs
dhvanipa: switched to using `get_json`, removed `get_text` entirely

cli/README.md
dhvanipa: updated!

SKILL.md
dhvanipa: Yes since the old ones are not deleted. But also this is probably not the release flow we want to go with to production. I was just manually uploading it to AWS lol, we should think about what we want there first, then implement something in follow up PR

dobjd/src/routes/events.rs
dhvanipa: yea in the current event data structure it shouldnt, but I changed this to an `.expect` so if we change it with a field that doesnt parse, it errors vs silently erroring

app-gui/src-tauri/src/objects/watcher.rs
dhvanipa: no, updates to the object collection happen on any progress events. So when the file is written during run action, changes to pending, changed to live, etc. all of these events trigger SSE events which updates the object collection. So no polling, the 4s polling is for `getGlobalStateRoot` to show that in the client

This makes it so if you drag an object from outside `~/.dobj/objects` into it, the GUI wont show you since its not polling. But I wasn't sure if this a clear user story yet, so I think we should visit it then

app-gui/src/shared/api/tauriClient.ts
dhvanipa: added timeout and error handling!

cli/src/commands.rs
```diff
@@ -0,0 +1,444 @@
...
+    let run_id = uuid::Uuid::new_v4().to_string();
+
+    // Subscribe to /events first so we don't miss progress messages emitted
+    // before the SSE connection is established. We block on the first `Open`
+    // event before posting the action so a fast `run_action` can't beat the
+    // EventSource handshake.
+    let events_url = format!("{}/events", client.base_url());
+    let progress_run_id = run_id.clone();
+    let (open_tx, open_rx) = oneshot::channel::<()>();
+    let progress_handle = tokio::spawn(async move {
```
ed255, line 169: Did you consider using an RPC library instead of rolling your own?
For "synchronous" methods I think your implementation of the http client with json is simple and serves our needs perfectly.
But for cases like this one, where you send a request and then wait for completion, perhaps using an existing solution could be beneficial.

ed255: I just noticed I originally misunderstood this code.
This is what I thought: you subscribe to an event channel, then send the request to run the action, which returns upon reception.  Then you wait until you see an event of the action being completed, and then you return.

But what's actually happening is that the events are only used to log the status of running the action.

The http post at `/actions/run` only completes once the action completes.  An action could be more than a minute, so I don't think this design is robust: an http post that takes more than a minute to return could be problematic, for example it could timeout at client side in regular configurations.  It would be annoying to have to tune every client to work around timeouts.

I would like to propose an alternative design where for rpc methods that have take a while first you send a request and get back an id.  And then you wait for the result via a mechanism like long polling, or subscribing to an event that informs the client when the request has completed.

cli/src/commands.rs
```diff
@@ -0,0 +1,444 @@
...
+    }
+    Ok(())
+}
+
+/// Stream every event flowing through dobjd's broadcast hub as JSON
+/// lines. Today this is `run-action-progress`.
+/// Each event prints once across all connected clients, so this is the
+/// single place to see activity from the desktop, web UI, MCP, and
+/// other CLI invocations at the same time.
+pub async fn events(client: &DobjdClient) -> Result<()> {
```
ed255, line 251: So currently the events only return logs right?
Do you have other ideas in mind for the future?

cli/src/daemon.rs
```diff
@@ -0,0 +1,426 @@
...
+        }
+        (None, _, false) => {
+            println!("dobjd is not running");
+            println!("  start with: dobj start");
+        }
+    }
+    Ok(())
+}
+
+pub async fn logs(follow: bool, lines: usize) -> Result<()> {
```
ed255, line 343: Do we assume Unix?  Or do we want this to also work on Windows?

Because if we only target Unix, `log` and `read_tail` could be simplified by using the `tail` command with `--follow` flag.

dobjd/src/progress.rs
```diff
@@ -0,0 +1,103 @@
...
+    run_id: String,
+}
+
+impl SseProgressReporter {
+    pub fn new(events: EventTx, run_id: String) -> Self {
+        Self { events, run_id }
+    }
+
+    fn send(&self, progress: RunActionProgress) {
+        let _ = self.events.send(Event::RunActionProgress(progress));
```
ed255: is it wise to ignore the error here?  could we at least log it?

relayer/build.rs
```diff
@@ -0,0 +1,68 @@
...
+        for entry in entries.flatten() {
+            let file_name = entry.file_name();
+            let Some(file_name) = file_name.to_str() else {
+                continue;
+            };
+            if !file_name.starts_with("scip-sys-") {
+                continue;
+            }
+
+            let candidate = entry.path().join("out/scip_install/lib");
```
ed255, line 23: I was wondering: why does the relayer need scip?  It's not going to use the MultiPodBuilder.
But I guess the answer is that it depends on pod2, and pod2 doesn't build without scip.

synchronizer/build.rs
ed255: how/why was this working before?
This is about the scip library link in `build.rs`, for crates that didn't have `build.rs` before.

wire-types/src/lib.rs
```diff
@@ -0,0 +1,418 @@
...
+/// `POST /actions/run` response.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[cfg_attr(feature = "schemars", derive(JsonSchema))]
+#[serde(rename_all = "camelCase")]
+pub struct RunActionResult {
+    /// Correlation id used for `run-action-progress` events. Echoed
+    /// from the request when supplied, otherwise a freshly-minted UUID v4.
+    pub run_id: String,
+    pub old_root: String,
+    pub new_root: String,
```
ed255, line 277: how about using `Hash` for all the fields that have `Hash` instead of `String` (in this and other structs)?
The benefits are:
- no need to manually convert, as this is handled by Serialize+Deserialize
- on the js side, this is encoded as a String
- If you don't like the current encoding of `Hash` we could change it in `pod2`

ed255 (commented):
I've done a high level review of this PR.  The main topic I would like to discuss is the design of rpc methods that take a long time to execute.
Aside from this, the rest looks good to me.

cli/src/commands.rs
dhvanipa: I agree that instead of blocking, a submit that returns right away would be more robust to timeouts and client/network crashes. I think this will require a small in-memory store though in the daemon to hold the `runId` state. so I will do this in a follow up PR

cli/src/commands.rs
dhvanipa: not really "logs" but right now its the events from any run action that is running. Some future ones I imagine are "inventory-changed", "settings-changed", "action-cancelled", etc.

cli/src/daemon.rs
dhvanipa: I dont think we should assume Unix, many users will be on Windows as well

relayer/build.rs
dhvanipa: Replying to the comment below which wont let me reply to it directly: "how/why was this working before?"

It was working before because we weren't running the synchronizer & relayer using the target release on the EC2 machine

dobjd/src/progress.rs
dhvanipa: The Err here is `tokio::sync::broadcast::SendError`, which fires if there are zero active receivers, which is the expected case when no UI/CLI is currently subscribed to `/events`. Logging at `warn!` would flood logs on every action when nothing's watching.

I added a trace log so its observable though

wire-types/src/lib.rs
dhvanipa: Agree we should do this, already being tracked here so will do in follow up PR: https://github.com/dobjlabs/zk-craft/issues/63

cli/src/commands.rs
dhvanipa: tracking here: https://github.com/dobjlabs/zk-craft/issues/101

dhvanipa: I think an interesting separate system to look at which has a similar thing is Ollama. Ollama is the most popular framework used to to run LLM inference locally. They have a very similar model of running an HTTP server locally that other clients call. 

For generating responses for prompts, they also have a HTTP endpoint, but instead of blocking POST/returning right away, they stream responses in the same HTTP request: https://docs.ollama.com/api/generate#generate-a-response

I think this works better for them though because you want to show user LLM response right away, vs for our action steps, there's not that may intermediate steps to show to the user

dobjd/src/progress.rs
ed255: ah, you already merged, but given your answer, I think it's fine to ignore the error.

PR #99, opened 2026-05-11
Action records

### LINKED ISSUE #513: StatementArg encoding with integer keys can allow collisions
`StatementArg::to_fields` encodes its three variants (`None`, `Literal(v)`, and `Key(root, key)`) into eight field elements with no discriminator. Two semantically different `StatementArg` values can produce identical field encodings, and the attacker chooses both sides. Because `MainPod::verify` binds only the statement hash to the proof's public input, a single proof verifies against either interpretation of the public statement.

`StatementArg::to_fields` at encodes the variants as:

```
None              -> [0, 0, 0, 0, 0, 0, 0, 0]
Literal(v)        -> [v.raw().0[0..4], 0, 0, 0, 0]
Key(root, key)    -> [root.0[0..4], key.to_fields()[0..4]]
```

The second half is zero for `Literal(v)` (padding) and also for `Key(root, IndexKey(0))`, because `IndexKey(0).to_fields()` is `[0, 0, 0, 0]`. That comes from `IndexKey::hash() = Hash::from(self.raw())` together with `RawValue::from(0_i64) = [0, 0, 0, 0]`.

Both of the following encode to the same eight field elements:

- `StatementArg::Key(r, IndexKey(0))`
- `StatementArg::Literal(v)` where `v.raw().0 == r.0`

The attacker controls `v`, so the match is constructible for any `r` they can observe. No grinding or hash inversion is needed.

`MainPod::verify` binds only the `sts_hash` (computed by flat-mapping `StatementArg::to_fields`) to the proof's public input. The same proof therefore verifies against either interpretation of the public statement: a verifier reconstructing the statement as `Key(r, IndexKey(0))` accepts it, and a verifier reconstructing it as `Literal(v)` with matching limbs also accepts it.

robknight:
This PR makes a few changes:

1. Instead of having one public wildcard per output, actions have two wildcards: one for all inputs, and another for all outputs
2. Instead of outputting one POD per object, we now output each object as a dictionary plus two Merkle proofs, approximately 1kb per file instead of 1mb
3. POD2 is bumped to a branch which has a better implementation of the predicate splitter and some bug-fixes to ensure that Records work

Currently each Record is padded so that the zero-index element is unused, until we can fix https://github.com/0xPARC/pod2/issues/513


sdk/src/lib.rs
dhvanipa, line 1: There are a bunch of changes here from the first SDK that @ed255 designed and implemented, so I think it'd be better if he reviewed this

txlib/src/predicates/crafting_test.podlang
```diff
@@ -11,18 +11,15 @@ use module 0xTXLIB_MODULE_HASH as tx
 // ========================================================
 
 DeleteLog(log, chain_start, chain_end) = AND(
-  DictContains(log, "type", @self_predicate(IsLog))
```
dhvanipa, line 14: why was the type constraint removed here?

dhvanipa: oh I see later it was moved to be an argument passed into the tx predicates

dhvanipa: Was this change done for performance reasons?

txlib/src/predicates/txlib.podlang
```diff
@@ -83,37 +91,61 @@ Nullify(nullifier, obj, private: obj_key_hash) = AND(
...
+// Head + recursive tail, specialized per head event type. Skips the
+// ReplayElement layer the head would otherwise need. For Insert and
+// Mutate the body of `ReplayInsert`/`ReplayMutate` is inlined here
+// (those predicates have room to fit alongside the recursive tail in
+// 5 sub-statements). To keep the wildcard count within the pod2 limit
+// of 8, head-event objects are packed into a small private dict and
+// accessed via anchored keys: `ins.new`/`ins.new_live` for Insert,
+// `pair.old`/`pair.new` for Mutate. ReplayDelete is too large to
+// inline so its Step variant just references the wrapping predicate.
+ReplayContentsStepInsert(before_tx, after_tx, before_chain, after_chain,
```
dhvanipa, line 127: Just for my understanding: is this change motivated by performance? It took me a long time to understand what the txlib predicates were doing to verify their correctness, and so every change I'm kind of "scared" of lol

But I think it makes sense when it leads to performance gains

Cargo.toml
```diff
@@ -16,9 +16,7 @@ members = [
 resolver = "2"
 
 [workspace.dependencies]
-# Branch `next/2026-03-30.a`: Includes:
-# - bump params wildcards to 9
-pod2 = { git = "https://github.com/0xPARC/pod2", rev = "448b6608fc0e39e644ee36435ad977c2d4e122bb", default-features = false, features = [
+pod2 = { git = "https://github.com/0xPARC/pod2", rev = "6129bd04ae14ee0f2baf6f71f77ba92a93cb523e", default-features = false, features = [
```
dhvanipa, line 23: The PR description says this is on a branch, can you add a comment here with the branch name? It's useful when debugging since with just the commit hash its not obvious which branch this is on

txlib/src/replay.rs
```diff
@@ -495,36 +785,70 @@ fn build_replay_mutate(
...
-    let op_dc = ctx
+/// Build a `ReplayContentsStepMutate` statement: the inlined body of
+/// `ReplayMutate` plus the recursive `ReplayContents` tail. `old` and
+/// `new` are packed into a `pair` dict so they share a single
+/// wildcard slot (keeps the predicate at 8 wildcards). The TxMutate
+/// and guard sub-statements are re-anchored to `pair.old` / `pair.new`.
+/// `lm`/`nl`/`nn` are supplied by the caller (which already needs them
+/// for the recursive tail).
+#[allow(clippy::too_many_arguments)]
+fn build_replay_step_mutate(
```
dhvanipa: holy massive method lol

txlib/src/lib.rs
```diff
@@ -173,15 +173,65 @@ impl<'de> Deserialize<'de> for Tx {
...
+/// the source tx is not yet anchored.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct GroundingEvidence {
+    /// Source transaction commitment (== `source_tx.ctx.commitment()`).
+    pub tx_final: Hash,
+    /// Commitment of the source transaction's live set.
+    pub live_root: Hash,
+    /// Merkle proof for `DictContains(source_tx_ctx, "live", live_root)`.
+    pub live_in_tx_proof: MerkleProof,
```
dhvanipa, line 195: For a user to verify this object is valid they need verify this merkle proof and then check the tx is in the synchronizer tx set correct?

sdk/src/fmt_podlang.rs
```diff
@@ -41,11 +45,43 @@ fn fmt_var_at(name: &str, ts: usize, max_ts: usize) -> String {
     }
 }
 
+/// An action's chain max_ts must be at least this for the SDK to pack
+/// intermediate chain states into a `<Action>Chain` record. Below the
+/// threshold, the per-step scalar wildcards (`chain1`, `chain2`, ...)
+/// fit in fewer slots than the record-typed wildcard would cost.
+pub(crate) const CHAIN_PACK_MIN_TS: usize = 3;
```
dhvanipa, line 60: its nice this is configurable!

justfile
```diff
@@ -41,10 +41,6 @@ test:
 test-ignored:
     cargo test --workspace --release -- --ignored --nocapture
 
-# Run the slow end-to-end proof test
-test-e2e:
```
dhvanipa, line 83: why was this removed?

txlib/src/predicates/crafting_test.podlang
robknight: Partly! Mostly I noticed that in long predicates we have lots of this:

```
  DictContains(in.oil, "type", @self_predicate(IsOil))
  tx::TxDelete(chain2, chain1, in.oil)
  DictContains(in.water, "type", @self_predicate(IsWater))
  tx::TxDelete(chain3, chain2, in.water)
  DictContains(out.tar_a, "type", @self_predicate(IsTar))
  tx::TxInsert(chain4, chain3, out.tar_a)
```

We can collapse those down to a single line per object by doing the type-check inside insert/delete/mutate. This also removes a `DictContains` inside the replay steps, because we can reuse `TxInsert/Delete/Mutate` there.

txlib/src/predicates/txlib.podlang
robknight: I've added a README and some more detailed inline comments.

This change is for performance reasons - it's effectively a kind of loop unrolling, where the prover knows how many elements there are and what type they have, it can pick specialised predicates rather than having to dispatch through multiple levels. I think the new README and comments make this a bit clearer.

Cargo.toml
robknight: Done

txlib/src/lib.rs
robknight: Yes

justfile
robknight: I think this was a merge issue. I've reverted the change.

sdk/src/fmt_podlang.rs
```diff
@@ -41,11 +45,43 @@ fn fmt_var_at(name: &str, ts: usize, max_ts: usize) -> String {
...
+
+/// Schema name for an action's chain record (e.g. `LogToWoodChain`).
+pub(crate) fn chain_schema_name(action_name: &str) -> String {
+    format!("{action_name}Chain")
+}
+
+/// True if this action's chain has enough intermediate states to be
+/// worth packing into a record.
```
ed255: What happens if we never pack intermediate states into a record? Do we get too many wildcards such that the predicate splitter can't find a solution?

txlib/src/replay.rs
```diff
@@ -414,13 +582,86 @@ fn build_replay_insert(
...
+/// Build a `ReplayContentsStepInsert` statement: the inlined body of
+/// `ReplayInsert` plus the recursive `ReplayContents` tail. The `new`
+/// object and the resulting `new_live` set are packed into a tiny
+/// `ins` dict so they share a single wildcard slot (keeps the
+/// predicate at 8 wildcards). Body sub-statements anchor to
+/// `ins.new` / `ins.new_live` instead of using bare wildcards.
+/// `nl` is `live + {new}`, supplied by the caller (which already
+/// needs it for the recursive tail).
+#[allow(clippy::too_many_arguments)]
+fn build_replay_step_insert(
```
ed255: this function and the following ones have many arguments.
Would it make sense to declare a struct that contains the common args from all these functions?

ed255 (commented):
I've done an inital review pass in which I only reviewed `sdk` and `txlib`. 


I encountered the following optimizations that make the code more complex
- The chain intermediate states sometimes gets collapsed into a record.  I don't know if this is an optimization.  Because this doesn't always happens the code gets more complex.
- Sometimes the objects are used directly from `in.obj` or `out.obj`, other times they require an `ArrayContains` opening.

One reason why these points make the code more complex is that they are two ways of handing things, and the code implements both.  Moreover the same dynamic flows are done in `sdk/src/lib.rs` and `sdk/src/fmt_podlang.rs` (not fully, but some logics are duplicated)

sdk/src/fmt_podlang.rs
robknight: It can always find a solution, it just might have to split into more predicates than is optimal. This is the same move we made with the `in`/`out` wildcards, eliminating the "one wildcard per X" where X is unbounded.

txlib/src/replay.rs
robknight: I've refactored this around some structs which hold the values used during replay.

sdk/src/fmt_podlang.rs
robknight: I have tried to reduce the redundancy between `lib.rs` and `fmt_podlang.rs`.

driver/src/object_record.rs
```diff
@@ -18,7 +15,8 @@ pub enum ObjectStatus {
     Nullified,
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
 pub struct ObjectRecord {
```
ed255, line 11: This is not for this PR, but I think the code would be cleaner if `id` and `tx_hash` were `Hash` or `Value` instead of `String`.

txlib/src/replay.rs
```diff
@@ -6,688 +6,894 @@
...
+    ctx: &'a mut BuildContext,
+    stats: &'a mut TxStats,
+}
+
+/// Per-step immutable world view: the live/nullifier sets and the
+/// chain-scope bounds at the current point in the replay walk.
+#[derive(Clone, Copy)]
+pub(crate) struct ReplayFrame<'a> {
+    pub(crate) live: &'a Set,
+    pub(crate) nullifiers: &'a Set,
```
ed255, line 42: Just a quick note for the future: our current implementation of the Containers, which have a MerkleTree underneath, makes cloning very cheap.
For both memory and disk storage it's just copying an Arc and a Hash.

So if dealing with implicit lifetimes gets annoying consider cloning without any practical performance penalty (in this context)

txlib/src/replay.rs
```diff
@@ -6,688 +6,894 @@
...
+/// `btx` is the pre-mutate tx-context dict; `lm` is the live set with
+/// `old` removed; `nl` is `lm` with `new` inserted; `nn` is the
+/// nullifiers set with `nullifier(old)` accumulated. Owned because
+/// the caller also needs `nl`/`nn` to thread into the recursive tail
+/// frame.
+pub(crate) struct MutateScratch {
+    pub(crate) btx: Dictionary,
+    pub(crate) lm: Set,
+    pub(crate) nl: Set,
+    pub(crate) nn: Set,
```
ed255: These field names seem very mysterious to me.  You explain what they are in the comment above, but could you explain the reason behind these names so that the reader can infer the meaning from the name?

For example I guess in 
- `lm` the `l` is for `live`, but what is the `m` for?
- `nl` the `n` is for nullifier, but what about the `l`?
- `nn` the `n` is for nullifer, but what about the second `n`?

Alternative, if you come up with more descriptive names that would also help.

txlib/src/replay.rs
```diff
@@ -6,688 +6,894 @@
...
+        chain: Hash,
+        frame: ReplayFrame<'_>,
+    ) -> (Statement, Hash, Set, Set) {
+        assert!(
+            !events.is_empty(),
+            "build_replay_contents: empty event list (empty action scope is forbidden)"
+        );
+
+        if events.len() == 1 {
+            let (st_elem, c, l, n) = self.build_replay_element(&events[0], chain, frame);
```
ed255: Would this function (and all functions below) get too verbose by renaming:
- c -> chain
- l -> live
- n -> nulls
?

txlib/src/lib.rs
```diff
@@ -173,15 +173,65 @@ impl<'de> Deserialize<'de> for Tx {
...
+
+impl GroundingEvidence {
+    /// Construct from the source transaction's `ctx` dict and `live` set
+    /// (both available to the prover at finalize time) plus the produced
+    /// object whose membership we attest.
+    pub fn new(ctx: &Dictionary, live: &Set, obj: &Dictionary) -> anyhow::Result<Self> {
+        let tx_final = ctx.commitment();
+        let live_root = live.commitment();
+        let (_, live_in_tx_proof) = ctx.prove(&StrKey::from("live"))?;
+        let obj_in_live_proof = live.prove(&Value::from(obj.clone()))?;
```
ed255: nit: live is inside ctx right?
So you could skip passing `live` as arg and instead get it via:
```
let (live, live_in_tx_proof) = ctx.prove(&StrKey::from("live"))?;
let live = live.as_set().unwrap();
```

ed255 (approved):
Overall looks good to me!  I've left some small comments.

txlib/src/predicates/txlib.podlang
dhvanipa: thanks for the README, was helpful!

dhvanipa (approved):

PR #102, opened 2026-05-20
Remove blueprint, timelib, groth16

### LINKED ISSUE #11: Plonky2 vs Groth16 for MVP
https://github.com/dobjlabs/zk-craft/pull/5#discussion_r2877472760

> This code supports two modes:
> 
>     "groth16", the plonky2 proof is verified by a groth16 circuit and the groth16 proof is sent on-chain
>     "not(groth16)", the plonky2 proof is just shrunk and compressed and sent on-chain
> 
> In the MVP I don't think we need the two modes; we should pick one and remove the unused code.
> To answer this we need to figure out some details:
> 
>     Do we want a relayer that accumulates DO commitments?
>     What's the max number of commitments we want per transaction?

dhvanipa:
Removing a couple of things in the repo to reduce complexity (for self & agents):

1) blueprint is no longer needed because of the new type field
2) timelib is not needed currently because it is not planned for the MVP (we can always refer to it in git history)
3) groth16 is not needed currently because we want to release without it, so tactically we can add it later. It's also not in a working state currently anyways
4) removed the stale `app-gui/src-tauri/Cargo.lock`, there should only be one top level Cargo lock file


robknight (approved):
This looks good to me overall but I am unsure as to whether we want to get rid of the blueprint entry.

I don't think that there is a real security/correctness concern, there may be a UX concern - it's nice to have an entry which is just the string name of the object type. If we didn't have this, then we'd have to look it up from the Podlang predicate, but the names are not part of the module hash. And if I don't have the Podlang code/pexe file then any object with an unknown `type` entry is mysterious to me - I can't even search for it.

I think @ed255 may have had some related concerns. I suggest asking for his comments as well.

For avoiding agent confusion, I wonder if it makes sense to commit a CLAUDE.md/AGENTS.md file.

dhvanipa:
> This looks good to me overall but I am unsure as to whether we want to get rid of the blueprint entry.
> 
> I don't think that there is a real security/correctness concern, there may be a UX concern - it's nice to have an entry which is just the string name of the object type. If we didn't have this, then we'd have to look it up from the Podlang predicate, but the names are not part of the module hash. And if I don't have the Podlang code/pexe file then any object with an unknown `type` entry is mysterious to me - I can't even search for it.
> 
> I think @ed255 may have had some related concerns. I suggest asking for his comments as well.
> 
> For avoiding agent confusion, I wonder if it makes sense to commit a CLAUDE.md/AGENTS.md file.

I think the UX concern can be solved by showing the class name that is resolved from the pexe (dont need manual user lookups, system can do this). ie when you display a digital object in a UI, it shows "Wood" besides the type field but you can "inspect" to see the raw type if you're curious. I think if the pexe is not there it shouldnt show anything since you cant trust the file.

I think in general users shouldnt trust any object for which they dont know the type of, and a blueprint field would only mislead them to think that as long as they see "blueprint: wood", its the same wood they have.

And so, it being mysterious if you dont have a pexe to resolve the file is actually good in my opinion.

Re agent confusion, yes I think both claude & agent files would be good to add in general for this repo (better in a separate PR)

ed255:
Regarding the UX debate, I quite agree with @dhvanipa, perhaps it's better to not let the user rely on a blueprint field as it could be misleading; and if you don't know the recipe of the object the best thing you could do is ignore it until you learn the recipe.

On the other hand, having the blueprint could be very useful for debugging; because seeing a type hash is very opaque for humans.

But there's another point which is that a class is fully determined by the actions, and the actions are just a collection of relations.  If those relations are abstract there could be duplicate classes with aliased names.

For example consider module A which defines:
- apple: can be found
- apple juice: takes 2 apples
And module B which defines:
- water: can be found
- water bottle: takes 2 water

These two modules have the same predicates with different names.  Without blueprints, an apple and a water are the same.

ed255:
The removal of timelib and groth16 look good to me!  Removing unused parts is great :smile:
Since the blueprints are purely application layer and defined in the rhai with no consequences to the SDK or anywhere else; it would be fine to remove them now and bring them back if we consider them necessary.
On the other hand, given that removing them is very easy, we could just leave them for now?

dhvanipa:
> Regarding the UX debate, I quite agree with @dhvanipa, perhaps it's better to not let the user rely on a blueprint field as it could be misleading; and if you don't know the recipe of the object the best thing you could do is ignore it until you learn the recipe.
> 
> On the other hand, having the blueprint could be very useful for debugging; because seeing a type hash is very opaque for humans.
> 
> But there's another point which is that a class is fully determined by the actions, and the actions are just a collection of relations. If those relations are abstract there could be duplicate classes with aliased names.
> 
> For example consider module A which defines:
> 
> * apple: can be found
> * apple juice: takes 2 apples
>   And module B which defines:
> * water: can be found
> * water bottle: takes 2 water
> 
> These two modules have the same predicates with different names. Without blueprints, an apple and a water are the same.

For the example you gave, in my opinion, "apple" and "water" are actually the **same** objects (if they have same VDF/POW numbers). In digital objects, compared to other games, the definition **is** the object itself. And the labels are semantic meaning on top. If a game defines "apple" with X VDF, the definition of apple in their game is any object with that amount of work done. The "label" is not what makes it an apple.

This is similar to blockchain applications where a smart contract stores the value of a users inventory but not the name of it/graphic. In DUST, what made "lava" lava was that in the physics if a player was above object id X, they would lose health. This lava could be called "grass" in other clients or rendered as such, but that wouldn't change the physics. Similarly, the physics here are the statements themselves.

I think we should also encourage players to discover this fact. That oh, "let me try to cheat by making my own rocket object" but then discover they cant cuz they need the statements. And then they discover what the object predicate means, "oh its not the name that matters, its the hash of the statements".

dhvanipa:
> The removal of timelib and groth16 look good to me! Removing unused parts is great 😄 Since the blueprints are purely application layer and defined in the rhai with no consequences to the SDK or anywhere else; it would be fine to remove them now and bring them back if we consider them necessary. On the other hand, given that removing them is very easy, we could just leave them for now?

I think keeping it adds extra mental/conceptual overhead (for devs & users), and so I'm going to remove it for now.

PR #107, opened 2026-05-25
Provide a fast path for single-insert actions

robknight:
We have a lot of actions which only perform a single insert - "mining" or "gathering" actions for primitive resources like wood, water, etc.

To minimize the number of steps we need to take when replaying such an action for transaction verification purposes, we can provide a "fast path" which jumps directly to verifying the single insert and then moves on. This is no less safe than working downwards through `ReplayAction -> ReplayContents -> ReplayElement -> ReplayInsert`, but requires fewer custom predicate verifications. In practice this moves our mining/gathering transactions below the limit of 10 custom predicate verifications, allowing them to be completed in a single POD.


txlib/src/predicates/txlib.podlang
```diff
@@ -197,9 +197,16 @@ ReplayAction(before_tx, after_tx, before_chain, after_chain,
 // ReplayContents is still used *within* an action, where inserts,
 // mutates, deletes, and sub-actions can all appear. Empty top-level
 // transactions are forbidden by TxBuilder, so there is no Done leaf.
+//
+// `ReplayActionInsert` is a K=1 fast path: when the single top-level
+// action's body is a lone Insert, the entire transaction proves in
+// 2 cpvs from this OR (ReplayActions -> ReplayActionInsert), bypassing
```
dhvanipa: what is `cpvs`?

dhvanipa (approved):
lgtm

txlib/src/predicates/txlib.podlang
ed255: my guess: custom predicate verifications

robknight: Sorry, Custom Predicate Verifications. I'll remove the acronym.

ed255 (approved):
LGTM!

PR #112, opened 2026-05-26
Use a Record for state root in txlib

### LINKED ISSUE #100: State root can return to being a Merkle tree
When Merkle proofs were a scarce resource, we converted the state root to being a hash of three values. Instead, it could now be a small dictionary, which would reduce the need for custom predicates to open the dictionary in txlib (it would reduce the number of custom statements needed for verifying inputs by ~50%).

robknight:
Closes #100

Before we introduced cheap Merkle proofs for shallow Merkle trees, we reduced the number of Merkle proofs by implementing the state root as a hash of the transaction set, the nullifier set, and the past state roots array. This used `HashOf` statements and a custom predicate rather than `SetContains` and anchored keys to access the transaction set.

However, since we now have cheap Merkle proofs, we can return to the simpler approach, which uses fewer statements and eliminates a custom statement per input. This will also make it easier to extend the state root in future, e.g. by adding the created items set.

There is no change in the synchronizer or relayer, this is purely a change in how we expose the global state to txlib.

### DIFF
--- a/txlib/src/lib.rs
+++ b/txlib/src/lib.rs
@@ -34,7 +34,7 @@ use pod2::{
     frontend::{Operation, OperationArg},
     middleware::{
         EMPTY_VALUE, Hash, NativeOperation, OperationAux, OperationType, Statement, StrKey, Value,
-        containers::{Dictionary, Set},
+        containers::{Array, Dictionary, Set},
         hash_values,
     },
 };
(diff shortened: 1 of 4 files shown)


txlib/src/lib.rs
```diff
@@ -75,19 +75,21 @@ impl StateRoot {
         }
     }
 
-    /// 3-layer hash:
-    ///   H(H(txns_root, nullifiers_root), H(block_number, gsrs_root))
+    /// Merkleized dictionary view; its commitment is the canonical GSR.
+    /// Predicates access fields via anchored-key syntax (e.g.
+    /// `state_root.transactions`).
+    pub fn dict(&self) -> Dictionary {
```
ed255: What about using a record?  small merkle trees have 256 leafs with our current params, so the probability of finding a collision is significant even with a small number of entries.
A record would completely eliminate that.

ed255 (commented):
The change sounds good to me but I'd suggest using a record instead of a pure dictionary.

txlib/src/lib.rs
robknight: We have lots of dictionaries of this size. Most Digital Objects are dictionaries of this size. If we're concerned about collisions here then we should probably be concerned about collisions in those, too.

ed255: You're right... should we use records for all objects?  I'm afraid of moving in that direction because then an object state will be less self-describing, it would need to have an attached schema to give meaning to each object field.  Maybe we can delay this decision for a later time and keep using Dictionaries for objects and the state root for now.

dhvanipa (approved):
lgtm, but do we need to update downstream stuff? eg all other references to "state_root_hash"?

ed255 (approved):
LGTM! But I have similar questions to @dhvanipa, do we need to update the synchronizer as well?

robknight:
I don't think so - the synchronizer only sees the root hash because that's what gets posted publicly. The POD sees the full Merkle tree when making the proof. The change to the wildcard name in the Podlang code reflects the fact that the value is a record rather than an opaque hash.

txlib/src/lib.rs
```diff
@@ -75,22 +75,33 @@ impl StateRoot {
...
-            hash_values(&[Value::from(self.block_number), Value::from(self.gsrs_root)]);
-        hash_values(&[
-            Value::from(txn_nullifiers_hash),
-            Value::from(block_number_gsrs_hash),
+            Value::from(self.gsrs_root),
         ])
     }
+
+    /// Commitment of the state root array.
+    pub fn hash(&self) -> Hash {
```
ed255, line 93: The synchronizer calls this method to get the `new_gsr`:
https://github.com/dobjlabs/zk-craft/blob/8dd441529519b4385ef11258fa5018908a516e9a/synchronizer/src/state_machine.rs#L213-L219

So the synchronizer is already calculating the state root correctly in this PR

ed255: This, to me, confirms that no further change is required in the synchronizer.

PR #115, opened 2026-05-28
add driver import

dhvanipa:
When someone sends you a .dobj file, you need a way to import it into your inventory.

### DIFF
--- a/driver/src/driver.rs
+++ b/driver/src/driver.rs
@@ -19,7 +19,7 @@ use crate::execute::{
     build_relayer_payload, obj_type_hash, reconcile_objects, resolve_inputs, save_results,
     update_output_files, validate_execute_request,
 };
-use crate::object_record::parse_object_record_file;
+use crate::object_record::{ObjectRecord, parse_object_record_file};
 use crate::object_store::{
     ObjectFileEntry, ensure_store_dirs, load_object_files, matches_query, write_object_file,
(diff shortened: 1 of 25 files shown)


driver/src/driver.rs
```diff
@@ -337,6 +337,121 @@ impl Driver {
...
+    /// Import an external `.dobj` object — one not produced by this driver
+    /// (e.g. a file from outside `~/.dobj/`) — into local inventory.
+    ///
+    /// `dobj_json` is the raw JSON contents of a `.dobj` file. The object is
+    /// filed under a canonical name derived from its commitment — the file's
+    /// own `id` and filename are not trusted.
+    ///
+    /// Validation:
+    /// - the `class` must be one this driver's catalog knows, and the object's
+    ///   on-chain `obj["type"]` predicate hash must equal that class's hash
```
robknight: [NOTE] I get what this means but calling this an "on-chain" hash seems like a misunderstanding. The same term ("on-chain" is also used below). I think it might be better to say something like "committed" type predicate hash.

robknight (approved):
[NOTE] lgtm! There's a source comment which I think could be improved but otherwise everything looks fine

driver/src/driver.rs
dhvanipa: updated to say committed

PR #116, opened 2026-05-28
add dev remote

dhvanipa:
Add a dev remote command that makes it easy to use the local build against the deployed synchronizer & relayer


robknight (approved):
[NOTE] lgtm

PR #117, opened 2026-05-29
Identity based on initial object state

robknight:
This PR adds an `identity` entry to all newly-created objects. The identity is the hash of the object's initial state (excluding the `identity` entry itself).

`TxMutate` enforces that the identity does not change during mutation.

`TxInsert` now takes an extra argument, `initial`, which is the initial state of the object prior to the `identity` entry being inserted. Actions have a new `initials` private argument, which contains the initial states of any objects they create (this is not present in actions which do not perform any `TxInsert`s). The created objects which appear in the `out` record will include the `identity` entry.

### DIFF
--- a/plugins/craft-basics/manifest.toml
+++ b/plugins/craft-basics/manifest.toml
@@ -2,7 +2,7 @@
 name = "craft-basics"
 version = "0.1.0"
 # Rewritten by `cargo run -p pexe -- build plugins/craft-basics`.
-module_hash = "0c1b18862f9f9dac2576e8712df821abd828699e88bbe6b8bca84a0898a41516"
+module_hash = "b1209099905ee349660880077b4d1cb34f4296d2913716a0688b6f4b43112a30"
 
 [[classes]]
 name = "Log"
--- a/plugins/craft-basics/manifest.toml
+++ b/plugins/craft-basics/manifest.toml
@@ -2,7 +2,7 @@
 name = "craft-basics"
 version = "0.1.0"
 # Rewritten by `cargo run -p pexe -- build plugins/craft-basics`.
-module_hash = "b1209099905ee349660880077b4d1cb34f4296d2913716a0688b6f4b43112a30"
+module_hash = "cdb831c5a4f0087a6d3a93dee3583f94cda4a7fea90a163f90014356bfab9358"
 
 [[classes]]
 name = "Log"
--- a/plugins/craft-basics/manifest.toml
+++ b/plugins/craft-basics/manifest.toml
@@ -2,7 +2,7 @@
 name = "craft-basics"
 version = "0.1.0"
 # Rewritten by `cargo run -p pexe -- build plugins/craft-basics`.
-module_hash = "97ee84293bdec99d5dd41a1005e39a2cdefcead01ec6719dd8b9ce8c4d7c6c78"
+module_hash = "0c1b18862f9f9dac2576e8712df821abd828699e88bbe6b8bca84a0898a41516"
 
 [[classes]]
 name = "Log"
--- a/sdk/src/lib.rs
+++ b/sdk/src/lib.rs
@@ -542,12 +542,10 @@ impl ActionHandle {
         // dispatch side's array (Mutate contributes to both). Slot 0
         // is `_pad`; real entries start at slot 1 (see PAD_ENTRY).
         //
-        // For Output objects the script's view is pre-identity, but
-        // TxInsert stamps `new.identity = commitment(initial)` and
-        // the predicate references the stamped form via the `out`
-        // record. We cache the stamped dict once here so the
-        // out_array entry, the array_contains anchor below, and the
-        // post-event obj_dict downstream all see the same value.
+        // For Outputs, cache the identity-stamped dict once here so the
+        // three consumers that need the post-identity form -- the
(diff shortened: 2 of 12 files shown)


txlib/src/predicates/crafting_test.podlang
dhvanipa, line 1: Does this produce a new podlang for https://github.com/dobjlabs/zk-craft/tree/main/sdk#test-example? if so, we should update the README

plugins/craft-basics/manifest.toml
dhvanipa, line 1: I think there should be a diff on the episode-1 manifest `module_hash` as well right?

dhvanipa (commented):
We have an `id` field currently in the digital object file and a couple other structs

1) https://github.com/dobjlabs/zk-craft/blob/main/driver/src/object_record.rs#L12
2) https://github.com/dobjlabs/zk-craft/blob/main/wire-types/src/lib.rs#L156
3) https://github.com/dobjlabs/zk-craft/blob/main/wire-types/src/lib.rs#L172

We should either remove the field cuz now there's an id inside of the object fields, or set this top level id based on that

This id is also used to set the file name. In my opinion, the file name should rename the hash of the current object vs the permanent id but I'm curious to see what you think

If not done in this PR, then we should create an issue to track, cuz it will be confusing to have the new identity field and the current "id" in the codebase together

dhvanipa:
If your object identity got leaked, would people be able to figure out when your past updates happened? No right, cuz the live set hash is not the identity

sdk/src/lib.rs
dhvanipa, line 1: This is for the future, but we should in the SDK prevent people from setting the top level fields, eg `identity`. I think right now in the rhai script you can

dhvanipa:
Can you mint 2 objects with the same identity (by using the same key), and then update one of them a couple of times, and then try to update the other. I think no, because as soon as you start updating one, the previous state gets nullified

txlib/src/predicates/txlib.podlang
dhvanipa, line 1: If you require an object of a specific identity, do you need to check its type vs just looking at the value? I think not, since the identity is a a commitment of the initial state which includes the type field. ie changing the type would produce a different identity

dhvanipa:
I think now this identity capability lets us create a coin purse, where you can actually see which specific purse a coin came from, and even restrict that they can only be put back in that purse, correct?

dhvanipa (approved):
lgtm overall, excited to have identity! I didnt review sdk changes in detail and think it would be better if @ed255 reviewed that

ed255 (approved):
LGTM!

ed255:
With the addition of identity I would like to point out that we should come up with consistent terminology for the following:
- What does it mean for two objects to have the same identity?
- What does it mean for two objects to be equal?
- Should we distinguish the concept of "object" and "object state"?
  - For example, two objects are the same if they have the same identity, but they can be in different states?
  - We no longer say that an object can't be created/consumed twice, but instead that an object can't go through the same state twice

txlib/src/predicates/crafting_test.podlang
robknight: I've updated it

plugins/craft-basics/manifest.toml
robknight: Yes. fwiw I think having to update these hashes is quite tedious, and I wonder if we could automate this somehow.

sdk/src/lib.rs
robknight: You can't generate a proof which does this. `TxInsert` calls `DictInsert` to add the identity entry, which will fail if one already exists, and `TxMutate` prevents any change to it.

robknight:
> I think now this identity capability lets us create a coin purse, where you can actually see which specific purse a coin came from, and even restrict that they can only be put back in that purse, correct?

Yes (I'd have to think about how to do the coin purse specifically, but yes).

robknight:
> With the addition of identity I would like to point out that we should come up with consistent terminology for the following:
> 
> * What does it mean for two objects to have the same identity?
> * What does it mean for two objects to be equal?
> * Should we distinguish the concept of "object" and "object state"?
>   
>   * For example, two objects are the same if they have the same identity, but they can be in different states?
>   * We no longer say that an object can't be created/consumed twice, but instead that an object can't go through the same state twice

Agree, we need a word for "object state" (version, draft, state, etc)

We could also refer to the identity specifically as the "stable identity", to distinguish it from the "identity" of the specific state (the hash/commitment).

robknight:
> We have an `id` field currently in the digital object file and a couple other structs
> 
> 1. https://github.com/dobjlabs/zk-craft/blob/main/driver/src/object_record.rs#L12
> 2. https://github.com/dobjlabs/zk-craft/blob/main/wire-types/src/lib.rs#L156
> 3. https://github.com/dobjlabs/zk-craft/blob/main/wire-types/src/lib.rs#L172
> 
> We should either remove the field cuz now there's an id inside of the object fields, or set this top level id based on that
> 
> This id is also used to set the file name. In my opinion, the file name should rename the hash of the current object vs the permanent id but I'm curious to see what you think
> 
> If not done in this PR, then we should create an issue to track, cuz it will be confusing to have the new identity field and the current "id" in the codebase together

I'll create a new issue for this so that we don't force a re-review of this PR.

PR #119, opened 2026-05-29
support windows

dhvanipa:
- Add support for windows
- Move relayer & synchronizer types out of the crate so they can be used without importing the full crates (eg RocksDB is heavy)
- Remove scip from all builds since it's now optional in pod2 and not enabled here
- Add `INSTALL.md` for manual install instructions
- Add CI action to smoke test end to end creating a digital object on multiple platforms (each runs its own local synchronizer & relayer): `dobj start → status → actions → run FindLog → inventory → stop`.

image


dhvanipa:
A note for the future, but I think Github Actions is a lot more expensive than other CI runners and so once we get to a proper release & CI stage, we should consider switching to something like Depot (https://depot.dev/)

robknight (approved):
[NOTE] lgtm!

I can't say that I've fully investigated all of the yaml files but if this works and it has been tested with Windows then it seems good to go.

PR #122, opened 2026-06-01
sdk: embed set kvs in action.output

ed255:
Instead of declaring the key-values than an output object must have via the `action.set` function, define them directly in `action.output`.
This simplifies the usage and removes error checks: for example, we don't need to check if the user is doing `set` on a mutate or input object. We were not doing these checks yet, but they were planned.

Before:
```
    var pick = action.output("WoodPick");
    pick.set([
        ["durability", 100]
    ]);
```
After:
```
    var pick = action.output("StonePick", [
        ["durability", 200]
    ]);
```

And the `Inst::Set` in the sdk has been removed.

### DIFF
--- a/driver/src/pexe_catalog.rs
+++ b/driver/src/pexe_catalog.rs
@@ -21,13 +21,13 @@ use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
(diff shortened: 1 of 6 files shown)


ed255:
Converted to draft to signal that this is not urgent and we can discuss the change later

PR #128, opened 2026-06-02
Rename object ID fields to content_hash

### LINKED ISSUE #123: Use object stable identity to populate the "id" field for serialized objects
As discussed in comments on https://github.com/dobjlabs/zk-craft/pull/117

robknight:
Closes #123

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -48,7 +48,7 @@ pub async fn inventory(client: &DobjdClient, json: bool) -> Result<()> {
                 &inventory
                     .iter()
                     .map(|o| serde_json::json!({
-                        "id": o.id, "fileName": o.file_name, "class": o.class,
+                        "id": o.content_hash, "fileName": o.file_name, "class": o.class,
                         "status": o.status, "txHash": o.tx_hash,
                     }))
                     .collect::<Vec<_>>()
@@ -67,7 +67,7 @@ pub async fn inventory(client: &DobjdClient, json: bool) -> Result<()> {
             obj.status,
             obj.emoji,
             obj.class,
-            short_hex(&obj.id),
+            short_hex(&obj.content_hash),
         );
     }
     Ok(())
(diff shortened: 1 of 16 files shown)


cli/src/commands.rs
```diff
@@ -46,7 +46,7 @@ pub async fn inventory(client: &DobjdClient, json: bool) -> Result<()> {
                 &inventory
                     .iter()
                     .map(|o| serde_json::json!({
-                        "id": o.id, "fileName": o.file_name, "class": o.class,
+                        "id": o.content_hash, "fileName": o.file_name, "class": o.class,
```
dhvanipa: Should the "id" key here also be renamed to "content_hash"

cli/src/commands.rs
```diff
@@ -278,14 +278,14 @@ pub async fn inspect_object(client: &DobjdClient, file_name: String, json: bool)
         println!(
             "{}",
             serde_json::to_string_pretty(&serde_json::json!({
-                "id": obj.id, "fileName": obj.file_name, "class": obj.class,
+                "id": obj.content_hash, "fileName": obj.file_name, "class": obj.class,
```
dhvanipa: same comment here, renaming the printed key from "id" to "content_hash"

driver/src/driver.rs
```diff
@@ -302,17 +302,17 @@ impl Driver {
             let expected_hash = decode_hash_hex(required.hash.as_str()).ok();
             let candidate = live_objects.iter().find(|entry| {
                 entry.record.class == required.class
-                    && !used_ids.contains(&entry.record.id)
+                    && !used_ids.contains(&entry.record.content_hash)
```
dhvanipa: It's confusing to me when I see a variable with `_ids` and then the field passed in is something that doesnt look like "id". 

I would suggest also renaming the variable here to to something else

dhvanipa (commented):
overall lgtm since its mostly renaming, but I'm seeing some inconsistency in some keys being renamed to "content_hash", while other places there's still a notion of "id"

"id" as in identifier and "identity" are 2 different things, but to me it's ambiguous when I see "id" around some code with an object if its the content hash being used as the "id" or the "identity" field

> A set of 10 different .dobjs files/objects in code that represent a wooden pick all have the same "identity" and different "content hashes".

This statement feels very unambiguous to me and does not introduce the notion of "id"

driver/src/driver.rs
robknight: Oh, I just did LSP-based renaming of the relevant fields. I'll have to hunt down the other uses of the ID terminology.

PR #129, opened 2026-06-02
Fix broken pexe inspect fixtures and test for future breakage

robknight:
`pexe inspect plan` generates fake objects to use as inputs. However, the fake object generation code did not keep up with the change to add an `identity` entry to real objects.

This PR fixes that bug, and adds a test which generates plans for all of `craft_basics` objects, which would surface any future breakage.

### DIFF
--- a/pexe/src/fixtures.rs
+++ b/pexe/src/fixtures.rs
@@ -188,13 +188,8 @@ mod tests {
         assert_eq!(typ.raw(), Value::from(class_hash).raw());
     }
 
-    /// Plan every manifest action against freshly minted inputs. This is
-    /// the guard against fixture drift: mint -> ground -> plan drives the
-    /// same `TxBuilder` path (insert, delete, mutate) that real chain
-    /// objects take, so a synthetic object whose shape no longer matches a
-    /// real post-insert object fails here. The mutate actions (UseWoodPick,
-    /// UseStonePick) are the ones that pin `old.identity == new.identity`,
-    /// so a missing identity stamp surfaces as a panic on those.
+    /// Plan every manifest action against freshly minted inputs, ensuring
+    /// that the synthetic objects are valid and preventing drift.
     #[test]
     fn every_action_plans_with_synthetic_inputs() {
         let module = load_craft_basics();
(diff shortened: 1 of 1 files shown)


pexe/src/fixtures.rs
```diff
@@ -184,6 +188,28 @@ mod tests {
         assert_eq!(typ.raw(), Value::from(class_hash).raw());
     }
 
+    /// Plan every manifest action against freshly minted inputs. This is
+    /// the guard against fixture drift: mint -> ground -> plan drives the
+    /// same `TxBuilder` path (insert, delete, mutate) that real chain
+    /// objects take, so a synthetic object whose shape no longer matches a
+    /// real post-insert object fails here. The mutate actions (UseWoodPick,
```
dhvanipa: I think the last line in this comment is not needed. The LLMs often add this when they're reasoning for the change, but dont realize that its not something you need in the code (happens to me a lot too)

pexe/src/fixtures.rs
```diff
@@ -184,6 +188,28 @@ mod tests {
...
+    fn every_action_plans_with_synthetic_inputs() {
+        let module = load_craft_basics();
+        for action in module.actions() {
+            let input_classes: Vec<String> =
+                action.total_inputs().map(|r| r.class.clone()).collect();
+            let minted = mint_classes(&module, &input_classes).unwrap();
+            let state = build_synthetic_state(&minted).unwrap();
+            let executor = module.executor(true, state.grounding_witness.clone());
+            executor
+                .plan_action(&action.name, state.spendable)
```
dhvanipa, line 203: Does this also now run the pod solver for each action? I dont think we had a test like that before but that seems good to have

dhvanipa (approved):

pexe/src/fixtures.rs
robknight: Yes, most of my time reviewing Claude output is spent removing these self-justifying comments!

dhvanipa: I wonder if there's a prompt/skill we can give to Claude so it doesnt do this

(I also imagine Anthropic eventually fixing this by adjusting their system prompt, but for some reason they havent so far)

pexe/src/fixtures.rs
robknight: This runs MultiPodBuilder to generate the POD layout, but doesn't generate the proof.

PR #130, opened 2026-06-02
Global created set

robknight:
`txlib` now produces a `TxFinalized` statement which includes the live set along with the nullifier set and tx hash. The synchronizer collects the live object hashes and adds them to a global set of created items. For reasons of collision-avoidance, this is materialized as a POD2 array.

Proving that an object is a valid input means proving that the object is in the `created` array for a given state root. `InputsGrounded` in `txlib` now uses that as evidence, rather than the previous approach which required proving the presence of a transaction first, and then the presence of the object in the transaction's live set. This simplifies `txlib`'s Podlang a bit.

`TxFinalized` still includes a tx hash. However, the synchronizer no longer makes Merkle proofs about transactions, and the transaction set is not Merkleized or referenced in the global state root hash (the `created` objects set replaces it).

Clients still ask the synchronizer to confirm the acceptance of a transaction, but this is done by checking for the inclusion of the created items and nullifiers, not by transaction hash.

### DIFF
--- a/common/src/payload.rs
+++ b/common/src/payload.rs
@@ -33,6 +33,31 @@ pub fn read_elems<const N: usize>(bytes: &mut impl Read) -> Result<[F; N]> {
     Ok(elems)
 }
 
+/// A list of hashes, length-prefixed by a single `u8` count: at most 255
+/// entries.
+fn write_hashes(buffer: &mut Vec<u8>, hashes: &[Hash]) {
+    assert!(hashes.len() <= 255);
+    buffer
+        .write_all(&(hashes.len() as u8).to_le_bytes())
+        .expect("vec write");
+    for hash in hashes {
+        write_elems(buffer, &hash.0);
+    }
+}
+
+fn read_hashes(bytes: &mut &[u8]) -> Result<Vec<Hash>> {
+    let len = {
+        let mut buffer = [0; 1];
+        bytes.read_exact(&mut buffer)?;
+        u8::from_le_bytes(buffer)
+    };
+    let mut hashes = Vec::with_capacity(len as usize);
+    for _ in 0..len {
+        hashes.push(Hash(read_elems(bytes)?));
+    }
+    Ok(hashes)
+}
+
 #[derive(Clone, Debug, Eq, PartialEq)]
 #[allow(clippy::large_enum_variant)]
 pub struct Payload {
@@ -41,6 +66,13 @@ pub struct Payload {
     pub tx_final: Hash,
     pub state_root_hash: Hash,
     pub nullifiers: Vec<Hash>,
+    /// Commitments of the objects this tx leaves live. The synchronizer
+    /// inserts each into its global created set, rejecting any that already
+    /// exists (the creation-uniqueness check, like a nullifier double-spend).
+    /// The list is also reconstructed into the `live` set that the public
+    /// TxFinalized statement commits to, so the proof only verifies against
+    /// these exact commitments -- the list cannot be forged.
+    pub live: Vec<Hash>,
 }
 
 const PAYLOAD_MAGIC: u16 = 0xd10b;
@@ -54,13 +86,8 @@ impl Payload {
         self.proof.write_bytes(&mut buffer);
         write_elems(&mut buffer, &self.tx_final.0);
         write_elems(&mut buffer, &self.state_root_hash.0);
-        assert!(self.nullifiers.len() <= 255);
-        buffer
-            .write_all(&(self.nullifiers.len() as u8).to_le_bytes())
-            .expect("vec write");
-        for nullifier in &self.nullifiers {
-            write_elems(&mut buffer, &nullifier.0);
-        }
+        write_hashes(&mut buffer, &self.nullifiers);
+        write_hashes(&mut buffer, &self.live);
         buffer
     }
 
@@ -79,20 +106,14 @@ impl Payload {
         bytes = &bytes[len..];
         let tx_final = Hash(read_elems(&mut bytes)?);
         let state_root_hash = Hash(read_elems(&mut bytes)?);
-        let nullifiers_len = {
-            let mut buffer = [0; 1];
-            bytes.read_exact(&mut buffer)?;
-            u8::from_le_bytes(buffer)
-        };
-        let mut nullifiers = Vec::with_capacity(nullifiers_len as usize);
-        for _ in 0..nullifiers_len {
-            nullifiers.push(Hash(read_elems(&mut bytes)?));
-        }
+        let nullifiers = read_hashes(&mut bytes)?;
+        let live = read_hashes(&mut bytes)?;
         Ok(Self {
             proof,
             tx_final,
             state_root_hash,
             nullifiers,
+            live,
         })
     }
 }
@@ -194,7 +215,7 @@ mod tests {
         let vds_root = vd_set.root();
 
         let input = r#"
-        TxnFinalized(state_root_hash, tx_final, nullifiers) = AND(
+        TxnFinalized(state_root_hash, tx_final, nullifiers, live) = AND(
             Equal(0, 0)
         )
         "#;
@@ -216,6 +237,13 @@ mod tests {
             let nullifiers_set = Value::from(Set::new(HashSet::from_iter(
                 nullifiers.iter().map(|h| Value::from(*h)),
             )));
+            let live = vec![
+                Hash(Value::from(10i64).raw().0),
+                Hash(Value::from(11i64).raw().0),
+            ];
+            let live_set = Value::from(Set::new(HashSet::from_iter(
+                live.iter().map(|h| Value::from(*h)),
+            )));
             let state_root = Value::from("dummy_state_root");
             let st0 = builder.priv_op(Operation::eq(0, 0)).unwrap();
             let st_txn_finalized = builder
@@ -225,6 +253,7 @@ mod tests {
                         (0, state_root.clone()),
                         (1, tx_final.clone()),
                         (2, nullifiers_set.clone()),
+                        (3, live_set.clone()),
                     ],
                     Operation::custom(pred.clone(), [st0]),
                 )
@@ -245,6 +274,7 @@ mod tests {
                 tx_final: Hash(tx_final.raw().0),
                 state_root_hash: Hash(state_root.raw().0),
                 nullifiers: nullifiers.clone(),
+                live: live.clone(),
             }
         };
 
@@ -258,12 +288,16 @@ mod tests {
         let nullifiers_set = Value::from(Set::new(HashSet::from_iter(
             payload.nullifiers.iter().map(|h| Value::from(*h)),
         )));
+        let live_set = Value::from(Set::new(HashSet::from_iter(
+            payload.live.iter().map(|h| Value::from(*h)),
+        )));
         let st = Statement::Custom(
             pred,
             vec![
                 Value::from(payload.state_root_hash).into(),
                 Value::from(payload.tx_final).into(),
                 nullifiers_set.into(),
+                live_set.into(),
             ],
         );
         println!("st: {st:?}");
--- a/common/src/proof.rs
+++ b/common/src/proof.rs
@@ -14,6 +14,15 @@ use pod2::{
     },
 };
 
+fn hashes_to_set(hashes: &[Hash]) -> Set {
+    Set::new(
+        hashes
+            .iter()
+            .map(|h| Value::from(*h))
+            .collect::<HashSet<_>>(),
+    )
+}
+
 /// Decodes and validates a raw blob payload.
 ///
 /// Returns `Ok(Some(payload))` when the blob contains a valid, cryptographically verified
@@ -28,7 +37,7 @@ pub trait BlobParser: Send + Sync {
 // MockBlobParser
 // ---------------------------------------------------------------------------
 
-/// Mock parser: accepts JSON `{ "tx_final": "0x...", "nullifiers": [...], "state_root_hash": "0x..." }`
+/// Mock parser: accepts JSON `{ "tx_final": "0x...", "nullifiers": [...], "live": [...], "state_root_hash": "0x..." }`
 /// Hash bytes serialized as lowercase hex strings.
 /// Returns a `Payload` with a dummy empty `PayloadProof` since no real proof
 /// is needed for mock testing.
@@ -40,6 +49,7 @@ pub struct MockBlobParser;
 struct MockProofJson {
     tx_final: String,
     nullifiers: Vec<String>,
+    live: Vec<String>,
     state_root_hash: String,
 }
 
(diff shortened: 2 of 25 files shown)


dhvanipa:
FYI - since this is a core change, I've added @ed255 as a reviewer as well

common/src/payload.rs
```diff
@@ -33,6 +33,33 @@ pub fn read_elems<const N: usize>(bytes: &mut impl Read) -> Result<[F; N]> {
     Ok(elems)
 }
 
+/// A list of hashes, length-prefixed by a single `u8` count: at most 255
+/// entries. A transaction's nullifier list (its consumed inputs) and live
+/// list (its outputs) are bounded by the action's input/output count, which
+/// is far below 255.
```
dhvanipa: just to double check my understanding: the action input/output is below 255, since it is specifically controlled by the pod2 mainpod circuit parameters

It shouldnt increase the public arguments cuz its just in/out now, but I assume there's an implicit limit still from # of merkle proofs?

common/src/payload.rs
```diff
@@ -41,6 +68,13 @@ pub struct Payload {
     pub tx_final: Hash,
     pub state_root_hash: Hash,
     pub nullifiers: Vec<Hash>,
+    /// Commitments of the objects this tx leaves live. The synchronizer
+    /// inserts each into its global created set, rejecting any that already
+    /// exists (the creation-uniqueness check, like a nullifier double-spend).
+    /// The list is also reconstructed into the `live` set that the public
+    /// TxFinalized statement commits to, so the proof only verifies against
+    /// these exact commitments -- the list cannot be forged.
+    pub live: Vec<Hash>,
```
dhvanipa, line 75: I agree with @ed255 's comment on Telegram that we should call this something other than live, as other "non-live" object commitments still live here (ie its append-only)

maybe: "commitments" or "created"

(can be in a future renaming PR though tracked by a github issue, doesnt have to block this design change PR)

common/src/proof.rs
```diff
@@ -40,6 +49,9 @@ pub struct MockBlobParser;
 struct MockProofJson {
     tx_final: String,
     nullifiers: Vec<String>,
+    // Optional so a mock fixture may omit it; absent means an empty live set.
```
dhvanipa: why is this new live field optional vs the previous ones like nullifiers is not?

common/src/payload.rs
dhvanipa: oh I see its call "created" later on in other files

synchronizer/src/state_machine.rs
```diff
@@ -136,49 +159,92 @@ impl StateMachine {
...
-        state.transactions.insert(&tx_value)?;
-        state.metadata.tx_count += 1;
+        for obj in &payload.live {
+            // The array is 1-indexed: slot 0 is left empty so no real object is
+            // ever grounded at `ArrayContains(created, 0, ..)`. Index 0's
+            // RawValue is [0,0,0,0], the all-zeros shape that collides with
+            // `None` / `IndexKey(0)` in pod2's StatementArg encoding -- the same
+            // hazard StateRoot's `_pad` slot avoids. `created_count` stays the
+            // true object count; the array slot is `created_count + 1`.
+            let index = state.metadata.created_count as usize + 1;
```
dhvanipa, line 173: this is because of the pod2 integer keys collision right? ie we can remove this + 1 once that's fixed?

synchronizer/src/state_machine.rs
```diff
@@ -136,49 +159,92 @@ impl StateMachine {
...
             nullifier_count = state.metadata.nullifier_count,
             "Validated blob state update in slot derivation"
         );
         Ok(())
     }
 
+    /// Record a rejection outcome for a tx (keyed by `tx_final`) so a client that
+    /// submitted it can poll the reason instead of only ever timing out.
+    ///
+    /// Never downgrades an already-accepted tx. A duplicate of an accepted tx has
```
dhvanipa: What should happen in the case of a re-org? Some tx results (ie commitments & nullifier hashes) would no longer be in the state

synchronizer/src/app_db.rs
```diff
@@ -63,26 +150,133 @@ impl AppDb {
...
+            return Ok(Some(index));
         }
+        self.created_scan(created, commitment)
     }
 
-    pub fn tx_exists_batch(&self, roots: &CanonicalRoots, tx_hashes: &[Hash]) -> Result<Vec<bool>> {
-        let txs = self.open_transactions(roots.transactions)?;
-        tx_hashes
+    /// Stale-hint fallback: scan the array for `commitment`, repairing the cache
+    /// to its true index when found. O(n); only reached when the cached index
```
dhvanipa: I understand we're using an array instead of set for created because we're worried about collisions given the size of merkle trees we're currently working with. 

But I assume we're hitting a performance issue then when generating proofs? ie `O(n)` scans now?

common/src/payload.rs
robknight: This is the live set for a single transaction, which is still called "live".

synchronizer/src/state_machine.rs
robknight: Correct!

synchronizer/src/app_db.rs
robknight: The index avoids scans. There may be a better way to do this, but I'm not sufficiently familiar with the synchronizer to propose a more complete refactor.

sdk/src/lib.rs
```diff
@@ -2098,21 +2098,15 @@ impl SdkModule {
 #[derive(Debug, Clone)]
 pub struct SpendableObject {
     pub obj: Dictionary,
```
dhvanipa, line 2100: if this is just one field now, do we need a struct for it?

driver/src/object_record.rs
```diff
@@ -21,9 +20,10 @@ pub struct ObjectRecord {
     pub tx_hash: Option<String>,
     /// Object payload dictionary
     pub obj: Dictionary,
-    /// Per-object grounding evidence: source tx commitment + Merkle
-    /// proofs anchoring the object inside that source tx's live set.
-    pub evidence: GroundingEvidence,
+    /// Commitment of the transaction that produced this object. Used only to
+    /// correlate with the relayer, resolving the current Ethereum tx hash
+    /// across fee-bump replacements.
+    pub tx_final: Hash,
```
dhvanipa, line 26: damn interesting, this means now a dobj file doesnt really carry any proofs at all! before it has zk proofs, then a merkle proof, and now its just: state + commitment

state transition = in pexe file, tied together with file via type hash
integrity & liveness = at run time checking whether the commitment exists in the synchronizer for a given block number

integrity is implicit in that if it exists, then it would have had a valid ZK proof in some blob

synchronizer/src/state_machine.rs
robknight: I've added an issue to follow this up: https://github.com/dobjlabs/zk-craft/issues/133

common/src/proof.rs
robknight: Oh, that was to avoid breaking some fixture data. I've updated the fixtures now.

common/src/payload.rs
robknight: This is afaik a pre-existing restriction on the size of the payload, not something that comes from the circuits. I've simplified the comment.

sdk/src/lib.rs
robknight: I think it still helps to distinguish object dictionaries from other dictionaries which may not be objects.

synchronizer/src/state_machine.rs
ed255: Reference for the fix: https://github.com/0xPARC/pod2/pull/520

robknight: I've created an issue for applying this fix across the zk-craft codebase: https://github.com/dobjlabs/zk-craft/issues/134

txlib/src/predicates/txlib.podlang
```diff
@@ -13,12 +13,12 @@
      hash step, updating the live/nullifier sets, and calling the
      affected object's type guard to authorize the event.
 
-  3. InputsGrounded -- proves each input object existed in a prior
-     finalized transaction referenced by the state root.
+  3. InputsGrounded -- proves each input object is present in the
+     global live set carried by the state root.
```
ed255: isn't this `created set` instead?

txlib/src/predicates/txlib.podlang
```diff
@@ -13,12 +13,12 @@
...
 
   4. TxFinalized -- the public entry point. Ties input grounding,
      chain seeding, and full replay together. Exposes
-     (state_root, tx_final, nullifiers) publicly.
+     (state_root, tx_final, nullifiers, live) publicly.
```
ed255, line 21: I know that this `live` is only for the transaction and is not the `created set` in the state.
But maybe we can also call it `created`?
The reason are as follows:
- The tx contains a nullifiers that is local, and then the synchronizer appends it to the state
- I think the reason to use created instead of live in the state can also be applied here.  An object can be created and then spent in the same tx right?  So once the tx is submitted, an entry in `live` could already be also consumed.

common/src/payload.rs
ed255: Based on my comment in `txlib.podlang` I propose to also call this `created`.  Specially considering than an object in this set could already be consumed.

sdk/src/lib.rs
```diff
@@ -2210,11 +2204,7 @@ impl Executor {
     fn new_builder(&self) -> MultiPodBuilder {
         MultiPodBuilder::new(&self.params, &self.vd_set)
     }
-    fn new_tx_builder(
-        &self,
-        ctx: &mut BuildContext,
-        inputs: &[(Dictionary, GroundingEvidence)],
-    ) -> TxBuilder {
+    fn new_tx_builder(&self, ctx: &mut BuildContext, inputs: &[Dictionary]) -> TxBuilder {
```
ed255, line 2207: here we pass the inputs without their index in the created array, this feels strange.

ed255: Ah never mind, the index is passed in `Executor::new` via `grounding_witness`

synchronizer/src/app_db.rs
```diff
@@ -63,22 +70,64 @@ impl AppDb {
...
-        match txs.prove(&value) {
-            Ok(proof) => Ok((true, proof)),
-            Err(_) => Ok((false, txs.prove_nonexistence(&value)?)),
+    /// Record an object commitment's position in the created `Array`.
+    ///
+    /// The cache is a plain `commitment -> index` map: not Merkleized, not part
+    /// of any committed root. It is only ever read as a hint -- every membership
+    /// or proof query cross-checks the index against the `Array` at the queried
+    /// root, so a stale entry (e.g. one left behind by an abandoned reorg branch)
+    /// resolves to "absent" rather than a wrong answer.
```
ed255, line 79: Do you mean that if there's a reorg, these keys will be rewritten?

If we can only treat this as a hit, it means that we can't use this to confirm that an object doesn't exist in the created set?

synchronizer/src/app_db.rs
```diff
@@ -63,22 +70,64 @@ impl AppDb {
...
+    ) -> Result<Vec<bool>> {
+        let created = self.open_created(roots.created)?;
+        obj_commitments
             .iter()
-            .map(|hash| {
-                txs.contains(&Value::from(*hash))
-                    .map_err(|err| anyhow!("{err}"))
+            .map(|commitment| match self.created_index_get(*commitment)? {
+                None => Ok(false),
+                Some(index) => Ok(created.get(index as usize)? == Some(Value::from(*commitment))),
```
ed255, line 130: this function appears to be using the created obj->index map to test for existence/non-existence.

I think this is fragile.
Consider the following scenario:
- Block a' is processed and creates objects at index 100, 101, 102
- Reorg, instead of block a' we get block a, which doesn't have any DO tx.  At this point we're storing some mappings from objects to index for non-existing objects.
- Now we call this method for non-existing objects 100, 101, 102.  Ok I see, they will be in the mapping but not in the `created` array, so this function will return [false, false, false]

So the scenario I presented is fine in this implementation.

I guess the invariant is the following:
- If an object is in the db at `created` array, then we have it's valid index.  If it's not at `created`, that object may exist or not in the mapping.
- As long as we always check the object in `created` after getting it's index from the hint mapping, we're fine.

ed255: Here's a related discussion: How do we detect that two synchronizers have diverged?
If they diverge on the `created` or `nullifiers` array, we just compare roots, which is very quick; and then we could start diffing the contents.
If they diverge on the `created_index` we don't have a single hash to compare, we need to diff the entire thing.  This could be inconvenient.

Upon further thinking: I don't think that not being able to easily compare `created_index` between two synchronizers is a big issue because it's just an auxiliary relation that is derived from the `created` array.
What I think can be inconvenient is that in the case of a reorg the `create_index` will get invalid/outdated entries that can make it harder to debug issues.
It would be great if there was an easy way to remove all invalid entries from `created_index` in the case of a reorg.  I don't have a particular proposal on how to achieve that though.

Note that this comment is not about correctness or soundness, it's just about making it easier to debug issues or understand what's happening if something fails.

synchronizer/src/state_machine.rs
```diff
@@ -21,12 +21,13 @@ pub const MAX_GSR_AGE_BLOCKS: i64 = 300;
...
 struct WorkingState {
     /// Mutable non-root metadata accumulated while deriving the slot.
     metadata: HeadMetadata,
-    /// Persistent transactions set opened from `head.roots.transactions`.
-    transactions: Set,
+    /// Persistent global created-object set (a pod2 `Array`) opened from
+    /// `head.roots.created`.
+    created: Array,
     /// Persistent nullifiers set opened from `head.roots.nullifiers`.
     nullifiers: Set,
```
ed255, line 32: along this review I was thinking that we also had a nullifiers array, but I see it's a set.

synchronizer/src/state_machine.rs
```diff
@@ -123,10 +128,24 @@ impl StateMachine {
...
+        let mut payload_created = HashSet::with_capacity(payload.live.len());
+        for obj in &payload.live {
+            if !payload_created.insert(*obj) {
+                warn!(
+                    slot,
+                    block_number, "Duplicate created object within payload; rejecting"
+                );
+                return Ok(());
+            }
+            if self.is_created(state, *obj)? {
```
ed255, line 142: Note to myself: here we're using the `created_index` as aux data, but we still validate against the `created` array.

synchronizer/src/app_db.rs
ed255: I answer myself:
- In the case of a reorg, the keys may be rewritten (if the same object is inserted at a different index) or left stale, if that object is never created in the future.
- We can't just use this mapping to validate existence, we always need to query ` created` as a source of truth.

ed255 (approved):
LGTM!
Before merging please consider these two major discussion:
- possibly rename `live` in the tx to `created`
- it would be neat if after a reorg, the `created_index` map got the stale entries removed

synchronizer/src/app_db.rs
```diff
@@ -27,6 +27,13 @@ fn value_key(raw: RawValue) -> Vec<u8> {
     k
 }
 
+fn created_index_key(commitment: Hash) -> Vec<u8> {
```
dhvanipa, line 30: Adding the comment from our call here: might be better to move this into the relational db keyed on block number so then re-org handling is easy and can be done in the same way it currently is

synchronizer/src/api.rs
dhvanipa, line 1: previously the synchronizer README was updated with the new API changes but I think it got reverted

dhvanipa (approved):
besides the comments I already added above, and the main thing being handling re-orgs, lgtm!

PR #132, opened 2026-06-02
Add CLAUDE.md

robknight:


dhvanipa (approved):
lgtm! this will evolve as the codebase evolves, but for every major PR we can just get claude to update README & claudemd

PR #136, opened 2026-06-03
remove database auto-creation

### LINKED ISSUE #61: In production we don't want the regular DB user to have DB creation privileges

dhvanipa:
The synchronizer and relayer no longer `CREATE DATABASE` at startup — provisioning moves to an idempotent `just ensure-db` recipe (and `bootstrap.sh` in prod (no changes, we were already doing this)), so the DB role no longer needs the `CREATEDB` privilege.

Fixes https://github.com/dobjlabs/zk-craft/issues/61


robknight (approved):
lgtm!

PR #137, opened 2026-06-03
fix ignored test_vdf_pod broken by pod2 statement-reveal API change

dhvanipa:
`test_vdf_pod` was panicking with `index out of bounds: len is 0` at `pub_statements()[0]`: the outer MainPod had zero public statements.

`just test-ignored` now passes

### DIFF
--- a/intro_pods/vdfpod/src/lib.rs
+++ b/intro_pods/vdfpod/src/lib.rs
@@ -830,7 +830,7 @@ mod tests {
         let mut main_pod_builder = frontend::MainPodBuilder::new(&params, vd_set);
         main_pod_builder.add_pod(main_vdf_pod.clone())?;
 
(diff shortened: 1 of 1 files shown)


intro_pods/vdfpod/src/lib.rs
```diff
@@ -830,7 +830,7 @@ mod tests {
         let mut main_pod_builder = frontend::MainPodBuilder::new(&params, vd_set);
         main_pod_builder.add_pod(main_vdf_pod.clone())?;
 
-        let _ = main_pod_builder.reveal(&main_vdf_pod.public_statements[0]);
+        main_pod_builder.open_input_st(true, 0, &main_vdf_pod.public_statements[0])?;
```
ed255, line 833: Before this fix, the `reveal` method was returning an error but it was being ignored by `let _ =`.
We should avoid doing `let _ = `!

ed255 (approved):
LGTM!

PR #138, opened 2026-06-03
bound pexe decompression to prevent zip-bomb DoS

### LINKED ISSUE #111: you can zip bomb a user with a malicious pexe

dhvanipa:
A malicious `.pexe` plugin could zip-bomb a user: `read_entry` decompressed zip entries via unbounded `read_to_string`, so a few-KB archive declaring a multi-GB entry would exhaust memory.

Fixes https://github.com/dobjlabs/zk-craft/issues/111

## Fix
Three bounds at the single unpack chokepoint, sized far above any real plugin (largest bundled entry ~29 KiB):
- **Per-entry decompression cap** (1 MiB) — `read_entry` now reads through `take(..)` and rejects overflow instead of growing the heap unbounded.
- **Entry-count cap** (16) — rejects crafted central directories in `unpack_raw`.
- **Compressed-file cap** (8 MiB) — new `read_pexe_file` reads through a capped reader (TOCTOU-free); wired into the driver scan and both CLI paths.

### DIFF
--- a/driver/src/pexe_catalog.rs
+++ b/driver/src/pexe_catalog.rs
@@ -354,8 +354,7 @@ fn discover_plugins(actions_dir: &Path) -> Result<Vec<Plugin>> {
 
     let mut plugins = Vec::with_capacity(entries.len());
     for path in entries {
-        let bytes = std::fs::read(&path)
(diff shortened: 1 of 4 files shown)


robknight (approved):
lgtm!

PR #140, opened 2026-06-04
Manage created index via Postgres for simpler reorg handling

robknight:
This builds on #130 by storing the indices into the `created` array in Postgres, making it easy to delete stale entries on reorg.

The most visible change is to `process_blob`, which is split into two stages. The most visible change is to process_blob, which is split into two stages. The first (parse_blobs) parses and verifies the incoming blob(s), revealing which created-object commitments are affected. Then we can query Postgres to see if any of these already exist, and pass that into the second stage (apply_payload), which validates and applies each blob.

There are some tests which exercise Postgres machinery, and which are ignored by default. I have run these tests locally, where they pass.

I think we are still assuming that the synchronizer is reset frequently, and so we don't handle any upgrade paths here.

### DIFF
--- a/synchronizer/src/api.rs
+++ b/synchronizer/src/api.rs
@@ -302,14 +302,14 @@ fn grounding_witness(
     object_commitments: &[Hash],
     indices: &HashMap<Hash, i64>,
 ) -> anyhow::Result<GroundingWitnessSnapshot> {
-    let proofs = app_db.prove_created_for(roots, object_commitments, indices)?;
+    let witnesses = app_db.prove_created_for(roots, object_commitments, indices)?;
     let object_proofs = object_commitments
         .iter()
-        .zip(proofs)
-        .map(|(commitment, proof)| ObjectMembershipProof {
+        .zip(witnesses)
+        .map(|(commitment, witness)| ObjectMembershipProof {
             commitment: *commitment,
-            present: proof.present,
-            witness: proof.witness,
+            present: witness.is_some(),
+            witness,
         })
         .collect();
     Ok(GroundingWitnessSnapshot { object_proofs })
--- a/synchronizer/src/api.rs
+++ b/synchronizer/src/api.rs
@@ -302,18 +302,16 @@ fn grounding_witness(
     object_commitments: &[Hash],
     indices: &HashMap<Hash, i64>,
 ) -> anyhow::Result<GroundingWitnessSnapshot> {
+    let proofs = app_db.prove_created_for(roots, object_commitments, indices)?;
     let object_proofs = object_commitments
         .iter()
-        .map(|commitment| {
-            let (present, witness) =
-                app_db.prove_created_at(roots, *commitment, indices.get(commitment).copied())?;
-            Ok(ObjectMembershipProof {
-                commitment: *commitment,
-                present,
-                witness,
-            })
+        .zip(proofs)
+        .map(|(commitment, proof)| ObjectMembershipProof {
+            commitment: *commitment,
+            present: proof.present,
+            witness: proof.witness,
         })
-        .collect::<anyhow::Result<Vec<_>>>()?;
+        .collect();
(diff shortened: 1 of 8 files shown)


synchronizer/src/api.rs
```diff
@@ -212,14 +212,23 @@ async fn post_grounding_witness(
     Json(body): Json<GroundingWitnessRequest>,
 ) -> Result<Json<GroundingWitnessResponse>, (StatusCode, String)> {
     ensure_hash_query_limit("objectCommitments", body.object_commitments.len())?;
-    let snapshot = load_current_snapshot(&app_state).await?;
```
dhvanipa, line 215: why was `load_current_snapshot` removed in 2 spots in this file but kept in 2 other places?

synchronizer/src/state_machine.rs
```diff
@@ -34,6 +34,42 @@ struct WorkingState {
...
+/// list and the lookup set cannot drift apart.
+///
+/// The lookup set answers within-slot duplicate detection: the persistent
+/// created index reflects only committed slots, so it cannot see objects added
+/// earlier in this same slot.
+#[derive(Default)]
+struct CreatedAdditions {
+    added: Vec<(Hash, i64)>,
+    seen: HashSet<Hash>,
+}
```
dhvanipa: are both fields needed here or could this be simplified to just a single `HashMap<Hash, i64>`

synchronizer/src/state_machine.rs
```diff
@@ -34,6 +34,42 @@ struct WorkingState {
...
+
+    fn contains(&self, commitment: &Hash) -> bool {
+        self.seen.contains(commitment)
+    }
+}
+
+/// One slot's derivation result: the candidate next canonical head plus the
+/// object commitments (with array indices) it appended to `created`. The caller
+/// persists the additions to the created index when committing the slot.
+pub struct DerivedSlot {
```
dhvanipa, line 48: doesnt this exact info already live in `ProcessedSlot`?

If so, can remove the duplicated fields in processed slot and just make it:

```
pub struct ProcessedSlot {
    pub slot: u32,
    pub block_root: B256,
    pub parent_root: B256,
    pub block_number: Option<u32>,
    pub is_empty: bool,
    pub derived: DerivedSlot,
}
```

synchronizer/src/api.rs
```diff
@@ -290,11 +300,13 @@ fn grounding_witness(
...
     roots: &CanonicalRoots,
     object_commitments: &[Hash],
+    indices: &HashMap<Hash, i64>,
 ) -> anyhow::Result<GroundingWitnessSnapshot> {
     let object_proofs = object_commitments
         .iter()
         .map(|commitment| {
-            let (present, witness) = app_db.prove_created(roots, *commitment)?;
+            let (present, witness) =
+                app_db.prove_created_at(roots, *commitment, indices.get(commitment).copied())?;
```
dhvanipa: This is re-opening the created `Array` once per commitment, but I see there's a new `created_exists_for` which opens it once outside of the loop. Can the same be done here?

dhvanipa (approved):
lgtm correctness wise! Added some comments for simplification & efficiency

synchronizer/src/api.rs
robknight: In two places we need the index and the head, and those were changed. The other two need only the head but not the index, so they were not changed.

synchronizer/src/state_machine.rs
robknight: Oh, yes, that's much simpler!

dhvanipa:
the synchronizer README should also be updated with the new changes (Claude should be able to do it)

synchronizer/src/state_machine.rs
robknight: Done. To avoid misleading names, I also made ProcessedSlot into an enum so that we never populate `DerivedSlot` for an empty slot.

PR #141, opened 2026-06-04
Bump to latest pod2, remove array padding and fix 0/EMPTY mismatches

### LINKED ISSUE #134: Remove array padding
Once https://github.com/0xPARC/pod2/pull/520 is merged, we can remove the padding of arrays which was used to prevent the use of index zero. This appears in all array and record types, used in the SDK, txlib, and in the synchronizer.

robknight:
This bumps zk-craft to the latest POD2, and removes the array padding mitigation we were using for the "zero index" bug. It also tidies up a now-incorrect assumption that 0 == EMPTY_VALUE - this is no longer true, and so we should use the literal `{}` in Podlang to equate to EMPTY_VALUE. In addition, intro PODs use the correct integer encoding.

Closes #134

### DIFF
--- a/intro_pods/lt_eq_u256_pod/src/lib.rs
+++ b/intro_pods/lt_eq_u256_pod/src/lib.rs
@@ -432,7 +432,7 @@ mod tests {
     };
     use pod2::{
         backends::plonky2::{basetypes::DEFAULT_VD_SET, recursion::circuit::hash_verifier_data},
-        middleware::hash_str,
+        middleware::{EMPTY_VALUE, hash_str},
     };
 
     use super::*;
@@ -556,7 +556,7 @@ mod tests {
         let params = Params::default();
         let vd_set = &*DEFAULT_VD_SET;
         let rhs = RawValue([F::ZERO, F::ZERO, F::ZERO, F(0x0020_0000_0000_0000u64)]);
-        let hash = RawValue([F::ZERO, F::ZERO, F::ZERO, F::ZERO]);
+        let hash = EMPTY_VALUE;
 
         let pod = LtEqU256Pod::new_boxed(&params, vd_set.clone(), hash, rhs)?;
         pod.verify()?;
--- a/intro_pods/lt_eq_u256_pod/src/lib.rs
+++ b/intro_pods/lt_eq_u256_pod/src/lib.rs
@@ -556,7 +556,7 @@ mod tests {
         let params = Params::default();
         let vd_set = &*DEFAULT_VD_SET;
         let rhs = RawValue([F::ZERO, F::ZERO, F::ZERO, F(0x0020_0000_0000_0000u64)]);
-        let hash = RawValue::from(0i64);
+        let hash = RawValue([F::ZERO, F::ZERO, F::ZERO, F::ZERO]);
 
         let pod = LtEqU256Pod::new_boxed(&params, vd_set.clone(), hash, rhs)?;
         pod.verify()?;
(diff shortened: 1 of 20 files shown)


sdk/src/lib.rs
```diff
@@ -721,12 +717,14 @@ impl ActionHandle {
...
-        // loop, once `tx_builder.{insert,mutate,delete}` advances the chain.
-        let mut chain_step_values: Vec<Value> = vec![Value::from(0_i64); chain_max_ts.max(1)];
+        // chain_step_values[ts - 1] holds the per-ts chain hash for each
+        // intermediate ts in 1..chain_max_ts (the final ts is the public
+        // `chain` wildcard, not stored here). SubAction values come from
+        // `st_sub` (baked in at Rhai time); Object values are filled later
+        // by the event_sts loop, once `tx_builder.{insert,mutate,delete}`
+        // advances the chain.
+        let mut chain_step_values: Vec<Value> =
+            vec![Value::from(0_i64); chain_max_ts.saturating_sub(1)];
```
dhvanipa, line 727: previous `vec![Value::from(0_i64)]` were switched to `Vec::new()`, which was this one kept?

robknight: This one is different because it isn't padding. It's a fixed-sized array, pre-filled with zeroes as a placeholder Value.

intro_pods/lt_eq_u256_pod/src/lib.rs
```diff
@@ -556,7 +556,7 @@ mod tests {
         let params = Params::default();
         let vd_set = &*DEFAULT_VD_SET;
         let rhs = RawValue([F::ZERO, F::ZERO, F::ZERO, F(0x0020_0000_0000_0000u64)]);
-        let hash = RawValue::from(0i64);
+        let hash = RawValue([F::ZERO, F::ZERO, F::ZERO, F::ZERO]);
```
dhvanipa: nit: can we import and use `EMPTY_VALUE` here?

txlib/src/lib.rs
dhvanipa: nit: stale comment should no longer say "padded"

sdk/src/fmt_podlang.rs
dhvanipa, line 1: can we also update the padded references (`_pad`) in SDK README

txlib/src/lib.rs
dhvanipa: rename 0 here to {}

txlib/src/predicates/txlib.podlang
dhvanipa: rename 0 here to {}

sdk/src/fmt_podlang.rs
dhvanipa: same with txlib README `chain_start = H(live_set, 0)`

dhvanipa (approved):
Added some comments for missed places for renames in comments but otherwise lgtm!

dhvanipa:
Not for this PR, but with the `Allow anchored keys in intro statements ` pod2 change, does this also mean now we can update SDK to allow this?

robknight:
> Not for this PR, but with the `Allow anchored keys in intro statements ` pod2 change, does this also mean now we can update SDK to allow this?

I think we would be using anchored keys for intro statements already, except for the pattern of adding a "work" entry to objects which use either VDFs or hash-mining; this forces the existence of a wildcard and removes the circumstance where we would use anchored keys instead. I believe this pattern can be removed, as the statement already does enough to prove that the work was done, without adding the entry to the object.

Issue #15, opened 2026-03-03
Warm caches on synchronizer startup


robknight:
https://github.com/dobjlabs/zk-craft/pull/5#discussion_r2877548548

PR #142, opened 2026-06-04
preload proof artifacts

### LINKED ISSUE #30: The proof artifacts are not preloaded so the first proof takes MUCH longer than subsequent ones (driver)
Options:
a) add loading screen
b) preload artifacts in bundle

dhvanipa:
Fixes https://github.com/dobjlabs/zk-craft/issues/30, https://github.com/dobjlabs/zk-craft/issues/15

The synchronizer & relayer already pre-load proof artifacts on startup by calling `ProofParser::new()`. This PR makes the driver also pre-load all proof artifacts the actions will use.

I also added some more logs to see various timings.

The synchronizer, relayer, and dobjd now write similar pod2 caches so they sometimes crash if they are running in parallel. Also, the desktop app shows empty if the dobjd is not up yet. So, I updated the dev flow so it launches them one after another. In production, this is not an issue as users only run dobjd and synchronizer/relayer also have crash-recovery mechanisms.

### DIFF
--- a/common/src/shrink.rs
+++ b/common/src/shrink.rs
@@ -196,20 +196,3 @@ pub fn cache_get_shrunk_main_pod_circuit_data(
     })
     .expect("cache ok")
 }
-
-/// Warm pod2's recursive MainPod circuit disk cache.
-///
-/// The first proof a process generates pulls the recursive MainPod circuit data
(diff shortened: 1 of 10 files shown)


robknight (approved):
[NOTE] lgtm!

PR #143, opened 2026-06-05
Split txlib into two modules

robknight:
To reduce instances of compatibility breakage, this splits the "events" predicates out of the main txlib. Because these are the only predicates that will be called directly from pexe actions, we reduce the risk of a change in txlib causing pexe file hashes to change.

This does not mean that txlib changes cause no issues: if the TxFinalized hash changes then synchronizers will need to know about the new hash, for example. However, fixing a small number of synchronizer instances is easier than fixing many distributed pexe files and object files whose contents refer to those pexe file hashes.


dhvanipa (approved):
lgtm!

PR #144, opened 2026-06-05
Public releases and automatic updates

robknight:
This PR adds public releases and automatic updates. Public releases are made via the `zk-craft-releases` repo, which can serve public downloads while this repo remains private (we will change this soon though).

The `zk-craft-releases` repo has a release script which performs the previously-manual setup steps.

`dobj` gains a `dobj update` command which will update to a newer version, if one is available.


.github/workflows/install-test.yml
```diff
@@ -89,11 +93,10 @@ jobs:
...
     permissions:
-      # `gh release download` here targets the DRAFT release created by
-      # release.yml. Draft releases are invisible to read-only tokens, so
-      # this needs write — `contents: read` 404s on a draft. Tighten to
-      # `read` only once releases are published before the test runs.
-      contents: write
+      # Artifacts are downloaded from the public releases repo using
+      # RELEASES_REPO_TOKEN (see the download step); the job's own token
+      # is not used for anything, so keep it read-only.
+      contents: read
```
dhvanipa, line 99: Did you trigger this action to test if it works? I remember for me it was failing before with read so I changed to write, but if read works it's better

.github/workflows/release.yml
```diff
@@ -40,14 +40,29 @@ env:
...
+  # RELEASES_REPO_TOKEN, a fine-grained PAT with contents read/write on
+  # that one repo. Publishing a release creates the tag in that repo
+  # (pointing at its README commit); the source tag lives only here.
+  RELEASES_REPO: dobjlabs/zk-craft-releases
+
+# Release uploads go to RELEASES_REPO via RELEASES_REPO_TOKEN, so the
+# workflow's own token only needs checkout access.
 permissions:
-  contents: write
+  contents: read
```
dhvanipa, line 65: oh wait does read work here now because of the token?

.github/release-notes.md.tpl
```diff
@@ -2,15 +2,21 @@
...
 ## Manual install
 
-See [SKILL.md](https://github.com/dobjlabs/zk-craft/blob/main/SKILL.md) for the underlying steps.
+See the [README](https://github.com/dobjlabs/zk-craft-releases#readme) for the underlying steps.
+
+## Upgrading
+
+Already installed? Run `dobj update` to upgrade in place to this release. It
+swaps `dobj`, `dobjd`, and `bitcraft-mcp-proxy` as a unit (atomic, with
+rollback) and leaves your plugins under `~/.dobj/actions/` untouched.
```
dhvanipa, line 19: It also leaves existing objects right?

cli/src/daemon.rs
```diff
@@ -58,11 +58,19 @@ pub struct DaemonPaths {
     pub log_file: PathBuf,
 }
 
+/// Resolve `~/.dobj` (without creating it). The single source for the dobj
+/// home directory, shared by the daemon lifecycle here and the updater
+/// (`crate::update`) so the launcher and the installer can't disagree on where
+/// the install lives.
+pub(crate) fn dobj_home() -> Result<PathBuf> {
```
dhvanipa, line 65: There is already `default_dobj_root` and `DOBJ_HOME_DIR` which does the same thing in `driver` crate, could that be imported?

SKILL.md
dhvanipa, line 1: Is this different from the README in zk-craft-releases? Is it meant to be the same?

wire-types/src/lib.rs
```diff
@@ -370,6 +370,23 @@ pub struct ObjectsDirInfo {
     pub path: String,
 }
 
+/// dobjd `/healthz` body. A superset of the relayer/synchronizer health shape:
```
dhvanipa, line 376: Nice! I can also imagine synchronizer & relayer both also reporting their version in the health response

README.md
dhvanipa, line 1: ty for renaming stuff here! I think even `bitcraft-mcp-proxy` could become `dobj-mcp-proxy` (can be in future PR, not blocking)

cli/src/update.rs
```diff
@@ -0,0 +1,894 @@
...
+    }
+
+    if CURRENT_TAG == "dev" {
+        bail!("this is a dev build; update it by rebuilding, not via dobj update");
+    }
+
+    // The user explicitly asked to update and will sit through the download
+    // anyway, so this resolve is bounded only against a fully stalled
+    // connection (RESOLVE_TIMEOUT), not a merely slow one.
+    let target_tag = match version {
```
dhvanipa, line 63: version is being accepted verbatim here, then used as a path component for staging.  `download_and_stage` deletes that path and with `--force` non-semver tags are allowed, so a tag like `../..` or an absolute path can escape `~/.dobj/staging` and trigger `remove_dir_all` on unintended directories...

cli/src/update.rs
```diff
@@ -0,0 +1,894 @@
...
+}
+
+struct SwapEntry {
+    installed: PathBuf,
+    backup: PathBuf,
+    had_previous: bool,
+    installed_new: bool,
+}
+
+/// Two-phase swap: first rename every current binary aside to `<name>.old`,
```
dhvanipa, line 641: does renaming the current running binary work on all OS, including Windows? Claude is telling me no

Would this update process also become simpler if one assumed that the process if stopped before updated? And update happening via a install script? Or is this standard update procedure

.github/workflows/install-test.yml
robknight: Yes, it worked for me

cli/src/update.rs
robknight: Good catch. Tags are now validated before use.

cli/src/update.rs
robknight: As far as I know, renaming does work on all OSes. Deleting or directly overwriting does not, which is why renaming is used instead.

cli/src/daemon.rs
robknight: The CLI deliberately avoids depending on `driver` to avoid pulling in a larger dependency tree, but I could change this if that's an acceptable tradeoff.

dhvanipa: oh yea true, no lets not add it then!

SKILL.md
robknight: SKILL.md describes the old flow, and I think we should either remove it or update it. I haven't really used agents for crafting much, so I'm not sure how useful the skill is in practice.

wire-types/src/lib.rs
robknight: I'll create an issue for this.

README.md
robknight: I'm assuming that renames of unchanged items are out of scope for this PR and will be picked up in the renaming/reorg PR when that happens.

SKILL.md
dhvanipa: I think we can remove it, since we can just give it the install README now (can happen in future PR when we test)

README.md
dhvanipa: yup!

dhvanipa (approved):

PR #145, opened 2026-06-05
Type hashes as pod2 `Hash` instead of `String` across the sync/relayer APIs, driver, and DB

### LINKED ISSUE #63: simplify synchronizer & driver API to use RawValue/Hash instead of String

dhvanipa:
Hashes now cross every service, driver, and storage boundary as a pod2 `Hash` instead of a hex `String`. They serialize via pod2's native serde on the wire (JSON) and as `BYTEA` in Postgres, which removes the ad-hoc `encode_hash_hex`/`decode_hash_hex` conversions that were scattered across the API handlers and HTTP clients.

#### Intentionally still `String`

- Ethereum `tx_hash` (keccak, not a Goldilocks `Hash`).
- Frontend-facing wire types (`ObjectSummary`, `RunActionResult`, class/action hashes) stay `String` + `0x` so `cli`/`mcp` remain pod2-free; `dobjd` encodes `Hash → "0x…"` at its route boundary.

Fixes https://github.com/dobjlabs/zk-craft/issues/63


ed255 (approved):
nice! LGTM

PR #146, opened 2026-06-05
Batch relayer tx-hash lookups and add CORS

dhvanipa:
Replaces the relayer's per-pending-object tx-hash lookup with a single batched `POST /api/v1/proofs/tx-hashes` endpoint (chunked at 256, server-capped), so `sync_inventory` makes one request instead of one per object. Adds permissive CORS to the relayer, matching the synchronizer.

### DIFF
--- a/driver/src/clients/relayer_client.rs
+++ b/driver/src/clients/relayer_client.rs
@@ -1,3 +1,4 @@
+use std::collections::HashMap;
 use std::time::{Duration, Instant};
 
 use anyhow::{Result, anyhow};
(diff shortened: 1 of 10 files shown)


driver/src/clients/relayer_client.rs
```diff
@@ -1,16 +1,22 @@
...
+    JobStatus, JobStatusResponse, SubmitProofRequest, SubmitProofResponse,
+    TxHashesByTxFinalRequest, TxHashesByTxFinalResponse,
 };
 
 pub const RELAYER_POLL_TIMEOUT_SECS: u64 = 180;
 pub const RELAYER_POLL_INTERVAL_MS: u64 = 1500;
 
+/// Per-request cap on `tx_finals` sent to the relayer's `tx-hashes` endpoint.
+/// MUST stay at or below the server's `MAX_TX_HASH_QUERY_ITEMS`.
+const TX_HASH_BATCH_LIMIT: usize = 256;
```
robknight, line 18: TX_HASH_BATCH_LIMIT, MAX_TX_HASH_QUERY_ITEMS, MEMBERSHIP_BATCH_LIMIT and MAX_HASH_QUERY_ITEMS seem to be the same constant. Should they be defined once?

robknight (approved):
lgtm! One minor comment about constants which could possibly be unified, but not urgent and might not even be correct.

driver/src/clients/relayer_client.rs
dhvanipa: no, the tx hash is for relayer queries and the hash_query is for synchronizer hashes. I think its good if both are independent so they can be changed on their own. Similarly the driver should have independent control of how many to batch at once (cant exceed the max but could be lower, eg if getting rate limited)

I renamed `MEMBERSHIP_BATCH_LIMIT` though since the synchronizer no longer calls it that

|   | Server | Driver client | 
| ------------- | ------------- | ------------- |
| relayer  | MAX_TX_HASH_QUERY_ITEMS  | TX_HASH_BATCH_LIMIT |
| synchronizer  | MAX_HASH_QUERY_ITEMS | HASH_BATCH_LIMIT |

Issue #32, opened 2026-03-16
"Global Valid State Root" and "Global Nullified State Root" show the same root


dhvanipa:
The prototype UI shows a "Global Valid State Root" and "Global Nullified State Root", but in the synchronizer there is a single global state root. Currently, we're rendering the same root for both for this reason, but the UI needs to be updated to show only one

PR #147, opened 2026-06-05
Rename state types: StateRoot→StateHeader, GSR→"state root", drop Canonical prefix

### LINKED ISSUE #135: Slight renaming proposal for state associated types
Sometimes I get a big confused by the naming of types related to the state:

For example we have
```
pub struct StateRoot {
    pub block_number: i64,
    pub transactions_root: Hash,
    pub nullifiers_root: Hash,
    pub gsrs_root: Hash,
}
```
But this is not the state root, this is more like the state header.  It's a header because it's shallow, it doesn't contain the nested containers, just their roots.  So I'd call this `StateHeader`.

Then we have `gsrs` which means `global state roots`.  Why is it called `global` here, but then we don't say `GlobalStateRoot`?  I would propose removing the `global` prefix, because this is simply the state that the synchronizers agree upon via consensus and clients rely on.
Related to this, could we just call it `state_roots_history` or similar?  I find more descriptive.

In the synchronizer we have some types prefixed with `Canonical` like `CanonicalHead` or `CanonicalRoots`.  I think there's no need to name this structs `Canonical`, because structurally there's no difference between a StateHead and a CanonicalStateHead.  A canonical state is just a state that is agreed to be canon by the consensus protocol.

So overall I'd use names like these:
```
pub struct State {
    pub block_number: i64,
    pub transactions: Array,
    pub nullifiers: Set,
    pub state_history: Array, // history of state roots
}
pub struct StateHeader {
    pub block_number: i64,
    pub transactions_root: Hash,
    pub nullifiers_root: Hash,
    pub state_history_root: Hash,
}
```

and similar names in the synchronizer `StateHeader`, `StateMetadata`, etc.

dhvanipa:
Clarifies the state-model vocabulary across txlib, the synchronizer, wire types, and the DB:

- `StateHeader` (was `txlib::StateRoot`) is the shallow struct of Merkle roots; its hash is the **state root** (was "GSR"/"global state root", e.g. `current_gsr → current_state_root`). The array of past roots is the **state history** (was "gsrs"/"gsr history").
- Drop the `Canonical` prefix in the synchronizer: `StateHead`, `StateRoots`, `StateMetadata`.
- Pure rename, no behavior change

Fixes https://github.com/dobjlabs/zk-craft/issues/135

### DIFF
--- a/app-gui/src/features/proof-runner/ProofRunnerPanel.tsx
+++ b/app-gui/src/features/proof-runner/ProofRunnerPanel.tsx
@@ -34,9 +34,9 @@ export function ProofRunnerPanel() {
     }
   }, [runActive, proof.action]);
 
-  const globalRootRaw = proof.stats.globalStateRoot?.trim() ?? "";
-  const globalRootDisplay = globalRootRaw
-    ? truncateDisplayHash(globalRootRaw)
+  const stateRootRaw = proof.stats.stateRoot?.trim() ?? "";
+  const globalRootDisplay = stateRootRaw
+    ? truncateDisplayHash(stateRootRaw)
     : "0x----...----";
 
   const formatCpuDuration = (totalSecs: number) => {
@@ -127,7 +127,7 @@ export function ProofRunnerPanel() {
               <span className="root-dot live" />
               <span className="root-label">Global Valid State Roots</span>
             </span>
-            <span className="root-hash" title={globalRootRaw || undefined}>
+            <span className="root-hash" title={stateRootRaw || undefined}>
               {globalRootDisplay}
             </span>
           </div>
@@ -136,7 +136,7 @@ export function ProofRunnerPanel() {
               <span className="root-dot nullified" />
               <span className="root-label">Global Nullified State Roots</span>
             </span>
-            <span className="root-hash" title={globalRootRaw || undefined}>
+            <span className="root-hash" title={stateRootRaw || undefined}>
               {globalRootDisplay}
             </span>
           </div>
--- a/synchronizer/src/head.rs
+++ b/synchronizer/src/head.rs
@@ -9,7 +9,7 @@ pub struct StateRoots {
     /// Root of the persistent spent-nullifiers set.
     pub nullifiers: Hash,
     /// Root of the prior-state root array committed inside `txlib::StateHeader`.
-    pub state_history: Hash,
+    pub prior_state_history: Hash,
     /// Root of the full state root history array after appending the current state root.
     pub next_state_history: Hash,
 }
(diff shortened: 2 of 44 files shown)


ed255 (approved):
LGTM!

app-gui/src/features/proof-runner/ProofRunnerPanel.tsx
```diff
@@ -127,7 +127,7 @@ export function ProofRunnerPanel() {
               <span className="root-dot live" />
               <span className="root-label">Global Valid State Roots</span>
             </span>
-            <span className="root-hash" title={globalRootRaw || undefined}>
+            <span className="root-hash" title={stateRootRaw || undefined}>
```
robknight, line 130: I know this was pre-existing, but "Global Valid State Roots" shows the same value as "Global Nullified State Roots". I'm not sure that this makes sense.

synchronizer/src/head.rs
```diff
@@ -1,103 +1,103 @@
...
     /// Root of the persistent global created-object set (a pod2 `Array`).
     pub created: Hash,
     /// Root of the persistent spent-nullifiers set.
     pub nullifiers: Hash,
-    /// Root of the prior-GSR array committed inside `txlib::StateRoot`.
-    pub state_root_gsrs: Hash,
-    /// Root of the full GSR history array after appending the current GSR.
-    pub gsr_history: Hash,
+    /// Root of the prior-state root array committed inside `txlib::StateHeader`.
+    pub state_history: Hash,
```
robknight: I think this would be _slightly_ clearer if it were called `prior_state_history` (matching the comment above it). But it's a tiny distinction.

robknight (approved):
lgtm! Two small nits but I don't mind if those are ignored.

app-gui/src/features/proof-runner/ProofRunnerPanel.tsx
dhvanipa: re-opened issue to track fixing it: https://github.com/dobjlabs/zk-craft/issues/32

synchronizer/src/head.rs
dhvanipa: renamed in https://github.com/dobjlabs/zk-craft/pull/147/changes/beb94a583ed97f671239b5b412879572949dfab4!

dhvanipa: renamed some other places to keep it consistent: https://github.com/dobjlabs/zk-craft/pull/147/commits/99c9db0c14af7e5f2ab1b8fde4792a56fd04dd21

even the `StateHeader` now makes it explicit that the state_history root is not the current one that includes created and nullifiers but a prior one

PR #148, opened 2026-06-08
Add install command to dobj CLI

robknight:
This adds a `dobj install file|url` command, allowing users to install pexe files easily.

The actual installation is performed by the daemon, which then reloads the catalog - this avoids the need to restart the daemon in order to pick up the new pexe file.

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -463,6 +463,44 @@ pub async fn feasibility(client: &DobjdClient, action_id: String, json: bool) ->
     Ok(())
 }
 
+/// Install a plugin from a local path or an http(s) URL. The bytes are sent to
+/// dobjd, which writes the `.pexe` into `~/.dobj/actions/` and hot-reloads the
+/// catalog so the plugin is usable immediately (no restart).
+pub async fn install(client: &DobjdClient, source: String, json: bool) -> Result<()> {
+    let bytes = read_or_download(&source).await?;
+    let plugin: String = client.post_bytes("/actions/install", bytes).await?;
+    if json {
+        println!(
+            "{}",
+            serde_json::to_string(&serde_json::json!({ "plugin": plugin }))?
+        );
+    } else {
+        println!("installed plugin '{plugin}' - run `dobj actions` to see what it added");
+    }
+    Ok(())
+}
+
+/// Resolve the install source: an http(s) URL is downloaded, anything else is
+/// read as a local file path. The byte cap is enforced authoritatively by
+/// dobjd (the request body limit), so there's no client-side limit here.
+async fn read_or_download(source: &str) -> Result<Vec<u8>> {
+    let is_url = reqwest::Url::parse(source)
+        .map(|u| matches!(u.scheme(), "http" | "https"))
+        .unwrap_or(false);
+    if is_url {
+        Ok(reqwest::get(source)
+            .await
+            .with_context(|| format!("GET {source}"))?
+            .error_for_status()
+            .with_context(|| format!("download failed: {source}"))?
+            .bytes()
+            .await?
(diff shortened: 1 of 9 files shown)


driver/src/driver.rs
```diff
@@ -111,6 +112,40 @@ impl Driver {
...
+    /// Validates the archive and plugin name first; if the reload fails, the
+    /// just-written file is removed so it can't break the next startup.
+    /// Returns the installed plugin name.
+    pub fn install_plugin(&self, bytes: &[u8]) -> Result<String> {
+        let (manifest, _script) = pexe::unpack(bytes).context("invalid .pexe archive")?;
+        let name = manifest.plugin.name.clone();
+        crate::pexe_catalog::validate_plugin_name(&name)?;
+        let path = pexe::install(bytes, &self.paths.actions_dir, &name)?;
+        if let Err(err) = self.reload_catalog() {
+            let _ = std::fs::remove_file(&path);
```
dhvanipa: wait, if it an install of a plugin fails, wont it remove the old one? Or is the idea that the file names have unique versions? even then if you hit install again and it fails, it could remove your old file. Maybe it errors if there's an existing plugin with same name?

cli/src/commands.rs
```diff
@@ -463,6 +463,44 @@ pub async fn feasibility(client: &DobjdClient, action_id: String, json: bool) ->
...
+    let is_url = reqwest::Url::parse(source)
+        .map(|u| matches!(u.scheme(), "http" | "https"))
+        .unwrap_or(false);
+    if is_url {
+        Ok(reqwest::get(source)
+            .await
+            .with_context(|| format!("GET {source}"))?
+            .error_for_status()
+            .with_context(|| format!("download failed: {source}"))?
+            .bytes()
```
dhvanipa: is there a way to limit the amount of bytes downloaded? Its unbounded right now right?

dobjd/src/routes/mod.rs
```diff
@@ -41,6 +42,14 @@ pub fn router(app_state: AppState) -> Router {
         )
         .route("/actions", get(actions::list_actions))
         .route("/actions/run", post(actions::run_action))
+        // Raw `.pexe` bytes; raise the body cap from axum's 2 MiB default to
+        // the pexe limit so larger plugins aren't rejected before the daemon
+        // can validate them.
+        .route(
+            "/actions/install",
+            post(actions::install_plugin)
```
dhvanipa, line 50: This is not for this PR but a comment in general, that because the driver is a process running locally, any other process can call methods on it, eg use objects, and now with this new command install any pexes

driver/src/driver.rs
```diff
@@ -497,7 +526,7 @@ impl Driver {
             .iter()
             .map(|input| input.record.spendable())
             .collect::<Vec<_>>();
-        let spendable_outputs = self.deps.catalog.execute_action(
+        let spendable_outputs = self.catalog().execute_action(
```
dhvanipa: the action is previously read from the catalog in line 502, then re-read here to execute. But now since the catalog can change at runtime, it might be just worth pinning it at the start of the function: `let catalog = self.catalog()`

driver/src/driver.rs
robknight: Good point. I've made this so that a failed install will not remove an existing plugin.

cli/src/commands.rs
robknight: I've added a limit which mirrors the limit in the `pexe` crate.

driver/src/driver.rs
robknight: Done

dhvanipa (approved):

PR #149, opened 2026-06-08
make action runs non-blocking and crash-recoverable

### LINKED ISSUE #101: change driver daemon's /actions/run to be non-blocking
Currently the driver daemon's `/actions/run` is blocking until the action completes. This is not robust to timeouts and client/network crashes. We should change this to return immediately with the runId and then have a separate endpoint to subscribe/poll

https://github.com/dobjlabs/zk-craft/pull/95#discussion_r3266898520

dhvanipa:
## Why
`POST /actions/run` held the connection open for the whole proof + commit pipeline (minutes) and kept no record of the run. A dropped connection — client crash, network blip, timeout — lost the result even though the work had already committed on-chain. Progress was a global SSE firehose with no replay, so reconnecting couldn't recover either.

Fixes https://github.com/dobjlabs/zk-craft/issues/101

## Design
Decouple doing the work from delivering the result. The run becomes a first-class, followable object:

- **`POST /actions/run` returns `202 {runId}` immediately** and runs the pipeline on a background supervisor task that records terminal state on every exit path, including panic (a run can't get stuck).
- **In-memory run registry** keeps status + progress log + result/error, retained for a TTL then reaped. Not persisted — that's enough to survive client/network crashes (the goal); on-chain state and local `.dobj` files reconcile via sync regardless.
- **Follow a run two ways**, both off that one source of truth: poll `GET /actions/runs/{id}`, or stream `GET /actions/runs/{id}/events` (replays buffered progress via `Last-Event-ID`, then tails to terminal).
- **Idempotent** by `runId`; registry shared by HTTP and MCP.

Also fixes the timeout bugs that made the old deadlines unenforceable (per-request reqwest timeouts on the relayer/sync clients; split the relayer's doubled budget) and adds a 60s request backstop on all non-SSE routes.

## Clients & compat
CLI, GUI, and MCP follow the new start-then-poll/stream flow; MCP gains a `get_run` tool.


dobjd/src/runs.rs
```diff
@@ -0,0 +1,551 @@
+//! In-memory registry of action runs.
```
dhvanipa, line 1: It didn't feel important to have this be persisted on-disk anywhere. But curious to see if anyone thinks otherwise

driver/src/clients/mod.rs
```diff
@@ -1,11 +1,34 @@
...
 pub(crate) use synchronizer_client::{HttpSynchronizerClient, SynchronizerClient};
 pub use synchronizer_client::{SYNCHRONIZER_POLL_INTERVAL_MS, SYNCHRONIZER_POLL_TIMEOUT_SECS};
 #[cfg(test)]
 pub(crate) use synchronizer_client::{SynchronizerHead, SynchronizerMembership};
+
+/// Per-request ceiling for a single relayer/synchronizer HTTP call. The poll
+/// loops only re-check their overall deadline between iterations, so without a
+/// per-request timeout one stalled `send()` blocks past the deadline forever.
+/// Bounding each attempt makes the loop deadlines actually enforceable.
+const HTTP_REQUEST_TIMEOUT_SECS: u64 = 30;
```
dhvanipa, line 21: These timeouts still fragile, so I'm looking forward to the synchronizer change where we have better tracking of what happened to an action

dobjd/src/runs.rs
ed255: This is an interesting question.  I think we should not make this information persistent on disk, because this is a task request on the dobjd, which is a daemon running in the same machine that is expected not to crash.

Let's consider some reasons why we may want it to be stored on disk:
- The dobjd often crashes
  - If it crashes after completing the task and before the client has polled the task status, storing it to disk would be useful
  - If it crashes while completing the task, we could have the dobjd resume the task.  I would find this more interesting, but now it's not just about storing the task but storing the state of the task so that the dobjd can pick it up and continue

I tasks are very long and we find that the probability of the dobjd crashing or getting stopped in the middle is high enough, I think we should consider the option of resuming tasks that were left in the middle.  This involves storing task state persistently.  And it sounds complex.

cli/src/commands.rs
```diff
@@ -180,95 +181,128 @@ pub async fn run(
...
+    // POST no longer holds the connection open for the whole pipeline.
+    let accepted: RunAccepted = client
+        .post_json(
             "/actions/run",
             &RunActionRequest {
                 input: RunActionInput {
                     action: action.clone(),
                     input_object_paths: input_paths,
-                    run_id: Some(run_id),
+                    run_id: None,
```
ed255: should this field be removed and we just let the daemon assign a run_id?

dobjd/README.md
```diff
@@ -36,34 +36,53 @@ dev, Tauri's webview for the desktop app).
...
+result/error in an in-memory registry. Follow a run either way:
+
+- **Poll** `GET /actions/runs/{run_id}` for the current state (the
+  disconnect-recovery path — a client that lost its connection re-reads the
+  outcome here).
+- **Stream** `GET /actions/runs/{run_id}/events`, which replays the buffered
+  progress (honoring `Last-Event-ID` on reconnect) then tails live events
+  until the run is terminal.
+
+A re-used `runId` is idempotent: it returns the existing run instead of
```
ed255: How is a `runId` reused?  Does this happen when the client sets the `runId` in the request?
- What happens if the client sets an existing `runId` in the request, but with a different request data than the corresponding run of `runId`?
- What happens if the client sets a non-existing `runId` in the request?
What's the use-case of "re-use" here?

cli/src/commands.rs
```diff
@@ -180,95 +181,128 @@ pub async fn run(
...
+                if !quiet {
+                    eprintln!("[{phase}/{status}] {message}");
+                }
+                let terminal = status == "failed" || (phase == "commit" && status == "done");
+                if terminal {
+                    es.close();
+                    break;
+                }
+            }
+            // A stream hiccup is not a run failure: stop streaming and fall
```
ed255: the events endpoint supports resume (via `last-event-id` in the header), but we're not doing that here in case of a hiccup.  Why?

dobjd/src/runs.rs
```diff
@@ -0,0 +1,609 @@
...
+        self.request == *request
+    }
+
+    pub fn status(&self) -> RunStatus {
+        self.inner.lock().unwrap().status
+    }
+
+    /// Append a non-terminal progress event and advance the status to match
+    /// its phase. No-op once the run is terminal.
+    fn push_progress(&self, progress: RunActionProgress) {
```
ed255, line 69: The following three methods have:
```
        if inner.finished_at.is_some() {
            return;
        }
```
But the methods are about updating the state of the run.  So wouldn't it be an implementation bug if we try to update the state of the run for a run that has already finished?
If that's the case, I think it would be better to do this to catch such bugs:
```
assert!(inner.finished_at.is_none());
```

dobjd/src/runs.rs
```diff
@@ -0,0 +1,609 @@
...
+    run_id: String,
+    action: QualifiedName,
+    input_objects: Vec<String>,
+) -> Result<RunAccepted, RunIdConflict> {
+    let request = RunRequest {
+        action: action.clone(),
+        input_objects: input_objects.clone(),
+    };
+    let entry = match registry.start(run_id.clone(), request) {
+        StartOutcome::Existing(status) => return Ok(RunAccepted { run_id, status }),
```
ed255: So there are two ways to use this:
- As a client I don't pass a run_id and get one which is generated as a random uuid
- As a client I pass a run_id
I don't understand why we have two different paths.  Could we only implement one to simplify the code?
Why would a client make a request with the same run_id and same request payload twice?  After the first request we can already query the status.

Since I don't understand the use-case of the client choosing the run_id and making a request with an already existing run_id, I would ask to consider not letting the client choose the run_id, choose it at random in the daemon, and remove all code that handles Existing and Conflict cases.  Perhaps there's a use-case that I'm missing so please let me know :)

ed255 (commented):
Overall looks good but I don't understand the use-case of repeated run_ids chosen by the user.  Depending on the answer of why that use-case is needed I would have suggestions to simplify the code slightly.

dobjd/src/runs.rs
```diff
@@ -0,0 +1,609 @@
...
+            new_root: common::encode_hash_hex(&result.new_root),
+            output_files: result.output_files.clone(),
+            nullified_files: result.nullified_files.clone(),
+        });
+    }
+
+    fn finish_failure(&self, message: String) {
+        let progress = RunActionProgress {
+            run_id: self.run_id.clone(),
+            phase: ExecutionPhase::Commit,
```
robknight: This hard-codes the phase to `Commit`, but a failure could occur at an earlier phase, right?

dobjd/src/routes/actions.rs
```diff
@@ -1,76 +1,133 @@
...
-                new_root: common::encode_hash_hex(&result.new_root),
-                output_files: result.output_files,
-                nullified_files: result.nullified_files,
-            })
+                next = index + 1;
+            }
+            if terminal {
+                return;
+            }
+            ticker.tick().await;
```
robknight: We could also select on tx.closed() here, and stop sending events in the case that the client disconnected.

dobjd/src/runs.rs
```diff
@@ -0,0 +1,609 @@
...
+        let entry = Arc::new(RunEntry::new(run_id.clone(), request));
+        runs.insert(run_id, entry.clone());
+        StartOutcome::Started(entry)
+    }
+
+    pub fn get(&self, run_id: &str) -> Option<Arc<RunEntry>> {
+        self.runs.read().unwrap().get(run_id).cloned()
+    }
+
+    /// Drop terminal runs older than [`RUN_RETENTION`].
```
robknight, line 184: I presume we're OK with successful runs returning 404 after they've been dropped - there's no reasonable scenario where a client will wait a long time before checking this?

dhvanipa:
> Overall looks good but I don't understand the use-case of repeated run_ids chosen by the user. Depending on the answer of why that use-case is needed I would have suggestions to simplify the code slightly.

The original reason for a client-chosen `run_id` was to correlate `/events` progress to a run before the (then-blocking) POST returned. Now the POST returns the runId immediately, so its not needed. I've updated it so the daemon now always mints the run_id, and removed all the extra code

cli/src/commands.rs
dhvanipa: yup, removed

dobjd/README.md
dhvanipa: removed

dobjd/src/runs.rs
dhvanipa: removed run_id and all this code

cli/src/commands.rs
dhvanipa: Good catch, it should be resumed here too! Updated it

dobjd/src/runs.rs
dhvanipa: yup, its better to catch such bugs, changed to assert!

dobjd/src/runs.rs
dhvanipa: yes good catch, updated it to use the correct phase

dobjd/src/runs.rs
dhvanipa: yes, I think this is related to the question above about storing the state on disk, vs in-memory (https://github.com/dobjlabs/zk-craft/pull/149#discussion_r3375995969). I think its fine to drop this, because the client can also check the outcome via the inventory

dobjd/src/routes/actions.rs
dhvanipa: good idea, updated it to select on it!

ed255 (approved):
LGTM!

PR #150, opened 2026-06-09
Add a blob archiver

### LINKED ISSUE #7: add service that's storing all digital object blobs (archiver)
Currently, we're not storing any blobs. This means, new synchronizers need to trust a derived state snapshot from existing synchronizers.

This is fine for the MVP, but we may want a service that's storing all digital object blobs, so that this trust assumption is removed in the future. (it can also be a config option in the synchronizer itself, eg "archive")

[Alchemy](https://www.alchemy.com/docs/chains/ethereum/ethereum-beacon-api-endpoints/ethereum-beacon-api-endpoints/v-1-beacon-blob-sidecars-block-id) returns me blobs from > 18 days, (tried ~200 days), but they don’t make a guarantee of providing it in the docs. [Quicknode](https://www.quicknode.com/docs/bnb-smart-chain/eth_getBlobSidecars) also doesn’t make a guarantee.

[BlockNative](https://docs.blocknative.com/data-archive/blob-archive) used to have a blob archive service, but it’s deprecated now.

[Blobscan](https://api.blobscan.com/#/) also has an API but doesn't look like a stable project. It's open source so anyone can run it though.

ed255:
This PR introduces the blob archiver, whose sole purpose is to sync with the beacon network and store blobs persistently filtered by the configured address.

The service only uses the filesystem for persistent storage (no database).  The implementation should be safe against crashes:
- After a block is processed we do the atomic operation of renaming `header.json.tmp` to `header.json`.  This means that if `header.json` exists, the block directory contains valid data.
- On start, if stale data is found from a crash, it is deleted in reverse order of creation, to guarantee that there are no left overs even if there's a crash during the deletion process.
- The disk io order is the following:
  - 1. Create a symlink at by_slot block entry to a by_root dir
  - 2. Create the by_root block dir
  - 3. Store the headers metadata as `header.json.tmp`
  - 4. Store the blobs
  - 5. Rename `header.json.tmp` to `header.json`
- The deletion order first recursively deletes the by_root block dir and then the by_slot block entry

The archiver reuses the ethereum beacon client code that the synchronizer uses, so I moved it to a new crate.


eth-clients/.env.example
dhvanipa, line 1: Why is there an `.env` needed in the `eth-clients` folder? Dont we already have one in the archiver and synchronizer?

blob-archiver/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "blob-archiver"
```
dhvanipa: can we call this just `archiver`, same with the folder name

blob-archiver/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+url = { version = "2.3.1", features = ["serde"] }
+reqwest = { version = "0.12.15", features = ["json"] }
+reqwest-eventsource = "0.6.0"
+backoff = { version = "0.4.0", features = ["tokio"] }
+chrono = "0.4.42"
+axum = { version = "0.8.6", features = ["macros"] }
+axum-extra = { version = "0.12.6", features = ["query"] }
+tower-http = { version = "0.6", features = ["cors"] }
+futures-util = "0.3.31"
+rocksdb = "0.23.0"
```
dhvanipa: Why does the archiver need rocksdb and sqlx? I would suggest only adding the minimal amount of packages to keep builds small

blob-archiver/README.md
```diff
@@ -0,0 +1,67 @@
+# Blob archiver
+
+A service that archives blobs filtered by destination address.
+
+The node connects to an Ethereum and Beacon RPC nodes to follow beacon blocks and donwnload the blobs that were sent via transactions if they match the configured destination address.  At the same time the node offers an http API to serve the stored blobs.
```
dhvanipa: typo in `donwnload` --> `downloads`

blob-archiver/src/config.rs
```diff
@@ -0,0 +1,43 @@
...
+    pub http_bind: SocketAddr,
+    pub init_slot: u32,
+    pub rpc_url: String,
+    pub beacon_url: String,
+    // Only store blobs from transactions with this destination
+    pub filter_address: Address,
+}
+
+pub fn load_config() -> Result<Config> {
+    let _ = dotenvy::from_filename("synchronizer/.env");
```
dhvanipa: I think this should be `archiver/.env`

blob-archiver/src/config.rs
```diff
@@ -0,0 +1,43 @@
...
+use alloy::primitives::Address;
+use anyhow::Result;
+
+const DEFAULT_HTTP_BIND: &str = "127.0.0.1:3001";
+
+#[derive(Debug)]
+pub struct Config {
+    pub blobs_path: String,
+    pub http_bind: SocketAddr,
+    pub init_slot: u32,
```
dhvanipa: can we call this the same as it is in synchronizer? `initial_start_slot`/`INITIAL_START_SLOT`

blob-archiver/src/main.rs
```diff
@@ -0,0 +1,108 @@
...
+        .map(|h| h.slot + 1)
+        .unwrap_or(node.config.init_slot);
+    loop {
+        debug!("checking slot {}", slot);
+        let some_beacon_block_header = if slot <= head.slot {
+            node.beacon_cli
+                .get_block_header(BlockId::Slot(slot))
+                .await?
+        } else {
+            // TODO: Be more fancy and replace this with a stream from an event subscription to
```
dhvanipa: There's stream code in the synchronizer that we could potentially move out and share here (not for this PR but in future)

blob-archiver/src/main.rs
dhvanipa, line 1: I love how simple this is lol

blob-archiver/README.md
```diff
@@ -0,0 +1,67 @@
...
+## API endpoints
+
+- `/healthz` -> `{ ok : true }`
+- `/header` -> `{ root: B256, parent_root: B256, slot: u32 }`
+  - Returns the latest synced beacon block metadata
+- `/config` -> `{ filter_address: Address }`
+  - Returns the filtering address
+- `/v1/beacon/blobs/{block_id}` with query array via `versioned_hashes` -> `{ data: Vec<Blob> }`
+  - Follows the same API as the [Ethereum Beacon Node
+    API](https://ethereum.github.io/beacon-APIs/#/Beacon/getBlobs).  This way
```
dhvanipa: I think `get_blobs` actually accepts any `BlockId`

From your link:
```
Block identifier. Can be one of: "head" (canonical head in node's view), "genesis", "finalized", <slot>, <hex encoded blockRoot with 0x prefix>.
```

But the current API only accepts a `B256` `block_id`

blob-archiver/src/main.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    let last_header = Arc::new(RwLock::new(None));
+    let api_state = api::ApiState {
+        config: Arc::new(api::Config {
+            filter_address: node.config.filter_address,
+        }),
+        store: Arc::new(node.store.clone()),
+        header: last_header.clone(),
+    };
+    tokio::spawn(run_api_server(api_state, node.config.http_bind));
```
dhvanipa: I think API server errors are getting dropped here, if the API server crashes, the entire archiver should crash imo

blob-archiver/README.md
```diff
@@ -0,0 +1,67 @@
...
+    the archiver can be use as a drop-in replacement of a beacon node for
+    getting blobs.
+
+## Data structure
+
+All data is stored in the file system (no database required) following this schema:
+
+- `slot_dir`: `{BLOBS_PATH}/by_slot/{slot_hi}/{slot_med}/{slot_lo}` symlink to corresponding `root_dir`
```
dhvanipa: Can you create the issue to track adding support for Windows?

blob-archiver/README.md
dhvanipa, line 1: Also 2 more issues:
- create an issue to verify KZG commitments for blob payloads we’re receiving in synchronizer 
- create an issue to download blobs from other archivers

blob-archiver/src/main.rs
```diff
@@ -0,0 +1,108 @@
...
+                continue;
+            }
+        };
+        if let Some(prev) = &prev_beacon_block_header {
+            if beacon_block_header.parent_root != prev.root {
+                info!(
+                    "reorg: slot {} ({}) has different parent than us",
+                    beacon_block_header.slot, beacon_block_header.root
+                );
+                slot = prev.slot;
```
dhvanipa: `slot` is set back to a previous slot on reorg detected but `prev_beacon_block_header` is not. This seems like a bug, because otherwise it will keep looping forever

blob-archiver/src/node.rs
```diff
@@ -0,0 +1,457 @@
...
+        let mut missing_blobs = Vec::new();
+        if !missing_vhs.is_empty() {
+            let blobs = self
+                .beacon_cli
+                .get_blobs(beacon_block_header.root.into(), &missing_vhs)
+                .await?;
+            assert_eq!(blobs.len(), missing_vhs.len());
+            debug!("got {} blobs from beacon_cli", blobs.len());
+            for (vh, blob) in zip_eq(missing_vhs, blobs) {
+                let index = block_kzg_blob_commitments
```
dhvanipa: To consider in future, but we could even have the archiver verify each blob's KZG commitment that it gets from the beacon RPC

blob-archiver/src/node.rs
```diff
@@ -0,0 +1,457 @@
...
+}
+
+#[derive(Clone)]
+pub struct Store {
+    pub blobs_path: String,
+}
+
+impl Store {
+    pub(crate) fn delete_block_data(&self, slot_path: &Path) -> Result<()> {
+        if !fs::exists(slot_path)? {
```
dhvanipa: **from Claude:**

A dangling slot symlink makes startup recovery loop forever. 

if `!fs::exists(slot_path)? { return Ok(()) }` — `std::fs::exists` follows symlinks, so a broken symlink reports false and `delete_block_data` returns early without unlinking it. 

`last_header` then re-selects that same dangling symlink as the highest entry every pass → infinite loop, archiver never starts. 

The dangling link is producible: `node.rs:289` creates the symlink before `node.rs:294` creates its target dir, so a crash in between leaves one.

Fix: use symlink_metadata/try_exists semantics that detect broken links, and create the symlink last.

blob-archiver/src/node.rs
```diff
@@ -0,0 +1,457 @@
...
+        // Create the path with the symlink to the yet to be created block dir, indexed by slot
+        let slot_path = self.store.slot_dir(beacon_block_header.slot);
+        let mut slot_mid_path = slot_path.clone();
+        slot_mid_path.pop();
+        create_dir_all(&slot_mid_path)?;
+        let root_dir_rel = root_dir_from(
+            ["..", "..", ".."].iter().collect::<PathBuf>().as_path(),
+            &beacon_block_header.root,
+        );
+        unix::fs::symlink(&root_dir_rel, slot_path)?;
```
dhvanipa: **From Claude:**

Reprocessing a slot errors out on an existing symlink. `unix::fs::symlink(...)` returns `AlreadyExists` if the `by_slot` link already exists (restart, or the recovery path). The `?` propagates out of the loop and exits the process.

Fix: remove-then-create, or ignore AlreadyExists.

dhvanipa (approved):
besides the comments, lgtm!

eth-clients/.env.example
ed255: uh, this was a slip over, this file doesn't belong there.

blob-archiver/src/config.rs
ed255: I'm very confused by this, I was running the archiver locally successfully... Ok I realized what's going on.
1. `dotenvy::from_filename` has the error ignored.
2. The `dotenvy` by default reads the `.env` file from the current directory, automatically.  I was starting the archiver from the `archiver` directory where I had the `.env`.

I'll fix this following your suggestion but I think we should consider removing that line and just relying on having a `.env` file in the directory where we start the process.  Alternatively we could have a proper configuration file that is passed by flag.

blob-archiver/README.md
ed255: Ah yes, for our use-case we only need to query by `block_root`, and that's what the archiver API supports.  I'll update the README to explain the limitations.

blob-archiver/src/main.rs
ed255: you're right!

blob-archiver/src/node.rs
ed255: I'm not sure this is an issue.  When we reach this point, the symlink shouldn't exist:
1. If we restart and the shutdown was "graceful", then we'll start with the next slot, and its symlink doesn't exist
2. If we restart and the shutdown was not "graceful" and we have a stale block directory, the symlink will be removed by `delete_block_data`

I'd like to keep this as is, and if we end up getting an `AlreadyExists` error we track the bug which caused it.

blob-archiver/src/node.rs
ed255: oh that's a very good catch, I didn't realize `fs::exists` would traverse symlinks.  I'll change this to use `fs::symlink_metadata`

eth-clients/.env.example
ed255: removed in https://github.com/dobjlabs/zk-craft/pull/150/commits/9232aa5afa22124494f335b8e4d8fce605d0d8b5

blob-archiver/Cargo.toml
ed255: renamed in https://github.com/dobjlabs/zk-craft/pull/150/commits/9232aa5afa22124494f335b8e4d8fce605d0d8b5

blob-archiver/Cargo.toml
ed255: Removed unused dependencies in https://github.com/dobjlabs/zk-craft/pull/150/commits/9232aa5afa22124494f335b8e4d8fce605d0d8b5

blob-archiver/README.md
ed255: renamed in https://github.com/dobjlabs/zk-craft/pull/150/commits/9232aa5afa22124494f335b8e4d8fce605d0d8b5

blob-archiver/src/config.rs
ed255: Renamed in https://github.com/dobjlabs/zk-craft/pull/150/commits/9232aa5afa22124494f335b8e4d8fce605d0d8b5
Opened this issue to discuss config files https://github.com/dobjlabs/zk-craft/issues/161

blob-archiver/src/config.rs
ed255: Renamed in https://github.com/dobjlabs/zk-craft/commit/9232aa5afa22124494f335b8e4d8fce605d0d8b5

blob-archiver/README.md
ed255: Updated the README in https://github.com/dobjlabs/zk-craft/commit/9232aa5afa22124494f335b8e4d8fce605d0d8b5

blob-archiver/src/main.rs
ed255: fixed in https://github.com/dobjlabs/zk-craft/commit/9232aa5afa22124494f335b8e4d8fce605d0d8b5

blob-archiver/src/node.rs
ed255: fixed in https://github.com/dobjlabs/zk-craft/commit/9232aa5afa22124494f335b8e4d8fce605d0d8b5

blob-archiver/src/main.rs
ed255: Actually :nerd_face: `run_api_server` calls `axum::serve` and the docs say:
> Although this future resolves to io::Result<()>, it will never actually complete or return an error. Errors on the TCP socket will be handled by sleeping for a short while (currently, one second).

https://docs.rs/axum/0.8.9/src/axum/serve/mod.rs.html#85-87

blob-archiver/src/main.rs
ed255: https://github.com/dobjlabs/zk-craft/issues/162

blob-archiver/README.md
ed255: https://github.com/dobjlabs/zk-craft/issues/163

blob-archiver/README.md
ed255: https://github.com/dobjlabs/zk-craft/issues/164
https://github.com/dobjlabs/zk-craft/issues/165

blob-archiver/src/node.rs
ed255: covered by https://github.com/dobjlabs/zk-craft/issues/164

PR #173, opened 2026-06-10
archiver: fix blob loss on restart and at the chain head

dhvanipa:
Fix archiver failure modes that can leave the synchronizer unable to fetch blobs from the archiver. The common production symptom is that the synchronizer requests a blob, gets a 404 from the archiver, retries 10x, and its sync loop dies.

All fixes are in `services/archiver/src/node.rs`.

## Fixes

1. **Resume walk dead-ended on relative `BLOBS_PATH`**

   `last_header()` descended the `by_slot` tree with `slot_path.push(entry.path())`. `DirEntry::path()` returns a full path, and `PathBuf::push` only replaces the buffer when that path is absolute.

   With a relative `BLOBS_PATH`, the walk appended paths like `blobs/by_slot/blobs/by_slot/...`, failed to find the last header, fell back to `INIT_START_SLOT`, and then tried to recreate an existing slot symlink.

   Fix: descend by adopting the entry directly with `slot_path = last.clone()`.

2. **Safe-delete guard rejected valid in-store paths**

   `delete_block_data()` compared a canonicalized block path against the raw `blobs_path`. A relative base or symlinked prefix such as macOS `/tmp -> /private/tmp` made the safety check reject legitimate paths.

   Fix: canonicalize the base path too, then compare canonical-to-canonical.

3. **Transient chain-head body miss committed a block as blobless**

   The main loop may fetch a slot header before the full beacon block body is available. The beacon client maps a 404 body lookup to `Ok(None)`, and the old code treated that as an empty slot, returned `Ok`, and committed `header.json` with no blobs.

   Fix: once a header exists, treat a missing full block body as an error instead of an empty slot. Docker restarts the process, resume cleanup removes the incomplete slot, and the archiver reprocesses it once the body is available.

   The genuinely empty slot path remains in the main loop where the header is missing. The genuine zero-blob block path is unchanged.

4. **Crash window could leave a dangling `by_slot` symlink**

   `process_beacon_block_header()` created the `by_slot` symlink before creating the corresponding `by_root` target directory. A crash in that small window left a dangling symlink that resume cleanup could not canonicalize.

   Fix: create the `by_root` directory before creating the `by_slot` symlink.

5. **Resume cleanup could stop at an empty highest slot subtree**

   A crash after creating a `by_slot` parent directory but before creating the slot symlink can leave an empty highest subtree, especially around a 1,000-slot boundary. `last_header()` would descend into that empty branch, conclude there was no stored header at all, and resume from `INIT_START_SLOT` instead of falling back to the previous valid slot.

   Fix: when `last_header()` finds an empty non-root `by_slot` directory, remove it and restart the descent so the walk can fall back to the next lower valid sibling. An empty root still means the store has no archived blocks.

## Follow-up

A better long-term fix is probably to make the archiver retry transient beacon inconsistencies the way the synchronizer does instead of relying on process restart. That needs a broader design discussion and is tracked in https://github.com/dobjlabs/digital-objects-network/issues/166.

### DIFF
--- a/services/archiver/src/node.rs
+++ b/services/archiver/src/node.rs
@@ -3,7 +3,7 @@ use std::time::Duration;
 use alloy::eips::eip4844::HeapBlob;
 use eth_clients::beacon::{
     self,
-    types::{BlockHeader, BlockId, KzgCommitment},
+    types::{Block, BlockHeader, BlockId, KzgCommitment},
     BeaconClient,
 };
 use itertools::zip_eq;
@@ -26,7 +26,7 @@ use alloy::{
 use anyhow::{anyhow, Context, Result};
 use backoff::ExponentialBackoffBuilder;
 use chrono::{DateTime, Utc};
-use tracing::{debug, info};
+use tracing::{debug, info, warn};
 
 use crate::config::Config;
 
@@ -92,9 +92,13 @@ impl Store {
             Err(e) if e.kind() == io::ErrorKind::NotFound => return Ok(()),
             Err(e) => return Err(e.into()),
         }
+        // Canonicalize the base too: blobs_path can be relative or sit under a
+        // symlinked prefix, so comparing it raw against a resolved path would
+        // reject legitimate in-store paths.
+        let blobs_root = fs::canonicalize(&self.blobs_path)?;
         let mut root_path = fs::canonicalize(slot_path)?; // Resolve symlink
         assert!(
-            root_path.starts_with(&self.blobs_path),
+            root_path.starts_with(&blobs_root),
             "root_path outside of blobs_path, removal is dangerous"
         );
         if fs::exists(&root_path)? {
@@ -143,8 +147,9 @@ impl Store {
                 } else {
                     break 'outer;
                 };
-                // Next level
-                slot_path.push(last);
+                // `entries` holds full paths, so descend by adopting the deepest
+                // directly; pushing a relative entry would append, not replace.
+                slot_path = last.clone();
             }
             // See if that block was completed, if not delete it and try again
             let mut header_path = slot_path.clone();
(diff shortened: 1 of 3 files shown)


services/archiver/src/node.rs
```diff
@@ -92,9 +92,13 @@ impl Store {
             Err(e) if e.kind() == io::ErrorKind::NotFound => return Ok(()),
             Err(e) => return Err(e.into()),
         }
+        // Canonicalize the base too: blobs_path can be relative or sit under a
```
ed255: Issue 2: :+1:

services/archiver/src/node.rs
```diff
@@ -138,13 +142,24 @@ impl Store {
...
+                    None => {
+                        // An empty directory below the root would otherwise
+                        // abandon the valid blocks under lower-numbered siblings:
+                        // drop it and retry so the descent falls back to one
+                        // instead of giving up. An empty root just means the
+                        // store holds no blocks.
+                        if level == 0 {
+                            break 'outer;
+                        }
+                        fs::remove_dir(&slot_path)?;
```
ed255, line 156: Issue 5: :+1:

services/archiver/src/node.rs
```diff
@@ -138,13 +142,24 @@ impl Store {
...
+                        continue 'outer;
+                    }
                 };
-                // Next level
-                slot_path.push(last);
+                // `entries` holds full paths, so descend by adopting the deepest
+                // directly; pushing a relative entry would append, not replace.
+                slot_path = last.clone();
```
ed255, line 162: Issue 1: :+1:

services/archiver/src/node.rs
```diff
@@ -274,7 +289,12 @@ impl Node {
         &self,
         beacon_block_header: &BlockHeader,
     ) -> Result<()> {
-        // Create the path with the symlink to the yet to be created block dir, indexed by slot
+        // Create the block dir where the filtered blobs and header will be stored, indexed by
```
ed255: Issue 4.

The original order of creating the symlink first was intentional.  The reason is that if we create the root dir first and the process crashes before creating the symlink we don't have an easy way to remove the root dir.

So having a dangling symlink was part of the original plan.  I would prefer it if we can find a way to make that work, so that we can do a proper clean up always.

services/archiver/src/node.rs
```diff
@@ -313,24 +329,19 @@ impl Node {
         let beacon_block_root = beacon_block_header.root;
         let slot = beacon_block_header.slot;
 
-        let beacon_block = match self
+        // We already hold this slot's header, so a None body is a transient miss
```
ed255, line 332: Issue 3: :+1:

services/archiver/src/node.rs
```diff
@@ -313,24 +329,19 @@ impl Node {
...
-            Some(payload) => payload,
-            None => {
-                debug!("slot {} has no execution payload", slot);
-                return Ok(());
-            }
-        };
+            .ok_or_else(|| {
+                anyhow!("Beacon header exists for slot {slot} but full block {beacon_block_root} was not found")
+            })?;
+        let execution_payload = beacon_block.execution_payload.ok_or_else(|| {
```
ed255: Issue 3 (part 2): I'm not sure about this. I was under the impression that a valid beacon block could have no execution payload.

ed255: I was mistaken: I've researched this and at some point in history the beacon block didn't have execution payload, but after "the merge" the execution payload became mandatory, so we won't find blocks today with no execution payload.  For this reason I think we should change the `eth-client` implementation so that `execution_payload` is not an `Option` anymore.

https://github.com/ethereum/consensus-specs/blob/v1.1.9/specs/bellatrix/beacon-chain.md#beaconblockbody

ed255 (commented):
I'll make a PR towards this branch with proposed changes for 2 of the issues

ed255:
Please take a look at this proposed amends to this PR https://github.com/dobjlabs/digital-objects-network/pull/185

ed255 (approved):
LGTM!  Note that I've merged my commit from https://github.com/dobjlabs/digital-objects-network/pull/185 into this branch already

PR #176, opened 2026-06-11
Add toggle for MCP server, off by default

### LINKED ISSUE #153: make MCP optional, add enable/disable

robknight:
Closes #153 

This adds a setting to enable/disable the MCP server. It's off by default, so must be enabled intentionally before use. Enabling/disabling takes effect immediately, so the MCP server will be stopped or started in response to the configuration change.

### DIFF
--- a/libs/wire-types/src/lib.rs
+++ b/libs/wire-types/src/lib.rs
@@ -386,13 +386,50 @@ pub struct RunActionProgress {
 // Driver configuration
 // ===========================================================================
 
-/// The driver's persisted configuration: synchronizer + relayer URLs.
+/// The persisted configuration shared by the driver and dobjd:
+/// synchronizer + relayer URLs, plus the daemon's MCP server toggle.
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
 #[cfg_attr(feature = "schemars", derive(JsonSchema))]
 #[serde(rename_all = "camelCase")]
 pub struct DriverSettings {
     pub synchronizer_api_url: String,
     pub relayer_api_url: String,
+    /// Whether dobjd serves MCP on the adjacent port (default off).
+    #[serde(default)]
+    pub mcp_enabled: bool,
+}
+
+/// Partial update for [`DriverSettings`], accepted by `PUT /settings`. Absent
+/// fields keep their current persisted value, so a caller can flip one
+/// setting without echoing the others. This is why the field type is
+/// `Option` rather than reusing `DriverSettings`: there, an omitted
+/// `mcpEnabled` would deserialize to `false` (`#[serde(default)]`, needed for
+/// reading older settings files) and silently stop the MCP server.
+#[derive(Debug, Clone, Default, Serialize, Deserialize, PartialEq, Eq)]
+#[cfg_attr(feature = "schemars", derive(JsonSchema))]
+#[serde(rename_all = "camelCase")]
+pub struct DriverSettingsPatch {
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub synchronizer_api_url: Option<String>,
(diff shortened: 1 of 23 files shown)


libs/wire-types/src/lib.rs
```diff
@@ -386,13 +386,50 @@ pub struct RunActionProgress {
...
+/// Partial update for [`DriverSettings`], accepted by `PUT /settings`. Absent
+/// fields keep their current persisted value, so a caller can flip one
+/// setting without echoing the others. This is why the field type is
+/// `Option` rather than reusing `DriverSettings`: there, an omitted
+/// `mcpEnabled` would deserialize to `false` (`#[serde(default)]`, needed for
+/// reading older settings files) and silently stop the MCP server.
+#[derive(Debug, Clone, Default, Serialize, Deserialize, PartialEq, Eq)]
+#[cfg_attr(feature = "schemars", derive(JsonSchema))]
+#[serde(rename_all = "camelCase")]
+pub struct DriverSettingsPatch {
```
dhvanipa, line 411: why is this needed vs wasnt needed before?

services/dobjd/src/main.rs
```diff
@@ -81,20 +100,6 @@ async fn main() -> Result<()> {
...
-    let mcp_event_tx = event_tx.clone();
-    let mcp_driver = driver.clone();
-    let mcp_runs = runs.clone();
-    tokio::spawn(async move {
-        if let Err(err) = start_mcp_server(mcp_driver, mcp_event_tx, mcp_runs, mcp_listener).await {
-            tracing::error!("MCP server crashed after startup: {err:#}");
-            std::process::exit(1);
-        }
-    });
-    tracing::info!("MCP server listening on http://{mcp_addr}/mcp");
```
dhvanipa, line 96: There's another place MCP is printed to the user and that is on `dobj start`:

https://github.com/dobjlabs/digital-objects-network/blob/d485db0d145edb08e54fc4bf3e22f2e457546455/interfaces/cli/src/daemon.rs#L402

Will need to check toggle there

services/dobjd/src/routes/settings.rs
```diff
@@ -11,8 +11,20 @@ pub async fn get_settings(State(state): State<AppState>) -> ApiResult<Json<Drive
...
+    let mut merged = state.driver.load_settings()?;
+    patch.apply_to(&mut merged);
+    // Reconcile the MCP server before persisting, so a failed enable (the
+    // adjacent port is taken) is reported to the caller without recording
+    // a setting that isn't in effect. If `apply` succeeds but the save below
+    // then fails, the running server and the persisted setting diverge with
+    // no rollback; save failures are rare and otherwise fatal, so we accept
+    // it rather than add a compensating stop/restart.
+    state.mcp.apply(merged.mcp_enabled).await?;
+    let saved = state.driver.save_settings(&merged)?;
```
dhvanipa, line 28: not for this PR, but we should create an issue to track which is that multiple driver calls writing to settings can race

eg both CLI & MCP go to set settings, and cuz there's no lock, only one change will get applied

dhvanipa (approved):
besides the comments, lgtm!

dhvanipa:
We should update the docs where the agent instructions for how to enable it

libs/wire-types/src/lib.rs
robknight: Before, the same DriverSettings struct was used for both the settings file and the PUT /settings request body. Adding `mcp_enabled` meant `#[serde(default)]` so existing config files (without the field) still parse, but that also made an API request omitting `mcpEnabled` deserialize to false instead of erroring, potentially disabling a running MCP server. The fix is a separate request type whose fields are all `Option` and get merged onto the current settings. DriverSettings stays as the file format and the API response.

PR #177, opened 2026-06-11
Return version/target info in health checks

### LINKED ISSUE #151: Services should report version numbers in health-check
`dobjd` now reports its version number in the health-check response, but other services should do so as well.

robknight:
Closes #151 

The synchronizer, relayer, and archiver now return version and target/platform information in health check responses.

One thing to consider: if there are known security issues with a particular version, this makes detecting the presence of the issue easier. We may want to restrict these routes to authenticated clients, or restrict the information available to authenticated clients in a follow-up.


dhvanipa (approved):
lgtm

dhvanipa:
> One thing to consider: if there are known security issues with a particular version, this makes detecting the presence of the issue easier. We may want to restrict these routes to authenticated clients, or restrict the information available to authenticated clients in a follow-up.

True, I assume we may want some auth in the future for other things like rate limiting as well

PR #178, opened 2026-06-11
MCP server: add a command UX (prompts), runtime commands, and a live dashboard

dhvanipa:
Three opt-in layers on top of the MCP server's generic tools — all in the `dobj-mcp` crate (the driver and dobjd are untouched), portable to any MCP client (no skills/hooks/client config). The UX is not enforced. You can still chat with Claude and just have it call tools, but you can also enter the commands UX via `/dobj:start`

### Prompts — a terse command UX
- `start` (dispatcher), `help`, `create-command`, `consult-docs`, and `dashboard` exposed as MCP prompts, plus one dynamic prompt per saved command.
- New `get_command` tool loads any command's full body (built-in or saved). The always-on server instructions tell the client to load and follow it when a command name is typed — so a command works typed inside a `start` session or invoked directly as `/mcp__dobj__<name>`.

### Runtime commands
- `define_command` / `list_commands` / `delete_command` tools.
- Stored as `~/.dobj/commands/<name>/README.md` (YAML frontmatter + instruction body) plus optional sibling scripts; `create-command` is the guided authoring flow.

### Live dashboard
- Self-contained `dashboard/index.html` (no build step), written to `~/.dobj/view/` on startup; the `view` prompt opens it as a Claude Code preview pane or hands back a URL. Polls the REST API for objects, the synchronizer head, and an action-log SSE.

Also adds `how-it-works` and `command-examples` reference docs (via `read_doc` and MCP resources) and refreshes the MCP README.

image


interfaces/mcp/dashboard/index.html
```diff
@@ -0,0 +1,638 @@
...
+      startEvents();
+
+      // Self-scheduling so a slow response never overlaps the next poll.
+      (async function loopUniverse() {
+        await pollUniverse();
+        setTimeout(loopUniverse, 5000);
+      })();
+      (async function loopObjects() {
+        await pollObjects();
+        setTimeout(loopObjects, 2000);
```
robknight: pollObjects() hits /objects which ultimately calls `Driver::sync_objects`, which will make requests to the relayer and synchronizer. Doing this every 2 seconds seems excessive.

interfaces/mcp/dashboard/index.html
```diff
@@ -0,0 +1,638 @@
...
+                `<div class="inv-row">` +
+                `<span class="name">${name}</span>` +
+                `<span class="badge ${status}">${escapeHtml(status)}</span>` +
+                `<span class="id">${id}</span>` +
+                `</div>` +
+                `</li>`
+              );
+            })
+            .join("");
+          card.innerHTML = `<ul class="list">${rows}</ul>`;
```
robknight: Before we get to this point, we could detect if there has been a change to the objects, so we don't have to rebuild the HTML every 2 seconds if there has been no change.

interfaces/mcp/src/server.rs
```diff
@@ -362,6 +477,40 @@ impl<T: DobjOps> ServerHandler for DobjMcpService<T> {
...
+    fn list_prompts(
+        &self,
+        _request: Option<rmcp::model::PaginatedRequestParams>,
+        _context: RequestContext<RoleServer>,
+    ) -> impl std::future::Future<Output = Result<ListPromptsResult, McpError>> + Send + '_ {
+        let mut prompts = crate::prompts::list();
+        if let Ok(store) = self.command_store() {
+            prompts.extend(store.list().iter().map(crate::prompts::command_prompt));
+        }
+        std::future::ready(Ok(ListPromptsResult::with_all_items(prompts)))
```
robknight, line 521: I wonder if we could cache this to avoid having to rebuild it frequently. Reading commands from disk could be cached by checking the `mtime` and not reloading if there has been no change.

robknight (approved):
Looks good overall, maybe some performance/caching optimizations and I think the risk of hitting the synchronizer too often might be worth fixing even if the others aren't.

interfaces/mcp/src/server.rs
dhvanipa: updated to be cached

interfaces/mcp/dashboard/index.html
dhvanipa: updated

interfaces/mcp/dashboard/index.html
dhvanipa: updated

PR #184, opened 2026-06-15
Include pexe binary in releases

robknight:
Includes the `pexe` binary in the bundle of binaries released via the installer.


dhvanipa (approved):
lgtm!

PR #188, opened 2026-06-16
Bump to latest POD2 with new argument order

robknight:
Updates to the latest POD2, which changes the name and argument order for some operations.


libs/pexe/src/inspect.rs
```diff
@@ -1444,7 +1444,9 @@ fn collect_fields_into_scope(
             | NativePredicate::ContainerUpdate
             | NativePredicate::DictInsert
             | NativePredicate::DictUpdate => {
-                (stmt.args.first(), stmt.args.get(2), stmt.args.get(3))
+                // New order (old, key, value, new): new root last,
+                // key/value in the middle, old root first (checked below).
+                (stmt.args.get(3), stmt.args.get(1), stmt.args.get(2))
```
dhvanipa, line 1449: why was it `1, 2, 3` before but now its `3, 1, 2`?

Cargo.toml
```diff
@@ -21,7 +21,7 @@ members = [
 resolver = "2"
 
 [workspace.dependencies]
-pod2 = { git = "https://github.com/0xPARC/pod2", rev = "17177e055b56595c29e2588707da324fad7f5f36", default-features = false, features = [
+pod2 = { git = "https://github.com/0xPARC/pod2", rev = "0f8ff85c6ec79b66513b905d6caef3f7029014ca", default-features = false, features = [
```
dhvanipa: I don't see the commit that corresponds to `0f8ff85c6ec79b66513b905d6caef3f7029014ca` on main

dhvanipa (approved):

Cargo.toml
robknight: Ah, it got a new number when it was merged. There was a bug in pod2 which is why I couldn't use the main branch at the time.

libs/pexe/src/inspect.rs
robknight: Because the old POD2 argument order was "new, old, key, value" and now it's "old, key, value, new".

ed255 (approved):
LGTM! I'm convinced this change was worth it.
Now native statements argument ordering is consistent with action arguments ordering, where we do (in, out) and not (out, it)

PR #190, opened 2026-06-17
fix archiver reorg off-by-one

### LINKED ISSUE #189: archiver infinite loop in reorg slot
Image

ed255:
The basic fix was an off-by one error.  We were deleting the slot yet to be stored, and not the previous one!
Since not all slots have blocks, there's a chance that in a reorg at slot n, slot n-1 has no block, so we're not deleting anything and we'd be stuck again.  So I've made a few changes to make sure we delete the previous slot, which may not be n-1 but lower.


dhvanipa (approved):
lgtm!

PR #197, opened 2026-07-15
fix tx events arg order

### LINKED ISSUE #196: argument order in tx_events.podlang is inconsistent
Predicates in `tx_events.podlang` use the order `after, before`, but we moved to `before, after` in https://github.com/dobjlabs/digital-objects-network/pull/188

ed255:
Resolve #196


robknight (approved):
lgtm!

PR #203, opened 2026-07-22
Add time and block hash to StateHeader

### LINKED ISSUE #200: Add time and block hash to StateHeader

ed255:
Resolve #200

Tested end to end with `just dev` and then `FindLog` from the gui.


libs/pod2utils/src/lib.rs
```diff
@@ -35,3 +36,16 @@ pub fn rand_raw_value() -> RawValue {
         }
     })
 }
+
+/// Lossy conversion of a 256 bit value to a pod2 Hash
+pub fn b256_to_hash(v: B256) -> Hash {
```
dhvanipa, line 41: hmm are there any implications of doing a lossy conversion? Would we ever need to recover the original 256-bit hash or verify a `StateHeader.block_hash` against a real Ethereum block hash?

libs/txlib/src/lib.rs
```diff
@@ -65,12 +70,16 @@ pub struct StateHeader {
 impl StateHeader {
     pub fn new(
         block_number: i64,
+        block_time: i64,
```
dhvanipa: nit: some places use `block_time` and others `block_timestamp`

eg podlang field uses `block_time` also

I would propose we use the same name everywhere, I'm okay with either

dhvanipa (approved):

libs/pod2utils/src/lib.rs
ed255: The implications of this lossy conversion is that it's only one way.  We can get a value that is derived from the block hash into digital objects.  But we don't have the original block hash, and we can't reliably recover the original block hash from it.  Right now for certain values (that appear with low probability) there will be multiple possible block hashes.  This should be fine to derive randomness in digital objects, but not to make statements about an ethereum block hash generally.

In the future we could instead calculate this value by hashing the original block hash with poseidon so that there's no multiple possible valid inputs; in that case supplying a witness will suffice to prove that the DO hash corresponds to a block hash.

libs/txlib/src/lib.rs
ed255: thanks.  I started using `block_time` but then I saw that in Ethereum world the term used is `timestamp` so I stared changing names but I see I forgot some of them.  I'll make sure to use `block_timestamp` everywhere.

libs/pod2utils/src/lib.rs
ed255: I've created an issue for this https://github.com/dobjlabs/digital-objects-network/issues/204

PR #205, opened 2026-07-29
Add support for state_header access in the SDK

### LINKED ISSUE #201: Extend SDK to read StateHeader ergonomically

ed255:
Resolve #201

I had to merge the In/Out records into a single IO one because even after the splitter did its thing, there were predicates that had 9 wildcards (our max is 8).
I believe this change doesn't introduce any performance regression and gets rid of one wildcard in some occasions.

Note for the reviewer: I haven't yet tested the correct behavior of the synchronizer after this PR nor an end 2 end flow because I don't have access to my everyday machine and I did this on 8GB of RAM :melting_face:


libs/sdk/src/fmt_podlang.rs
```diff
@@ -82,7 +82,7 @@ pub(crate) fn chain_schema_name(action_name: &str) -> String {
 /// record's array layout is `[step_0_value, step_1_value, ...]`, so the
 /// slot index is `ts - 1` and the step name is `step_{ts-1}`.
 pub(crate) fn chain_step_at(ts: usize, chain_max_ts: usize) -> Option<usize> {
-    (chain_packed(chain_max_ts) && ts > 0 && ts < chain_max_ts).then_some(ts - 1)
+    (chain_packed(chain_max_ts) && ts > 0 && ts < chain_max_ts).then(|| ts - 1)
```
ed255, line 88: There was an underflow here (which was detected in debug mode).
On underflow this results in None, but in debug an underflow panics.  By using a closure the subtraction only happens when the check passes.

dhvanipa:
what does `d--` in "kind": "d--" mean? I think d is dictionary but not sure what the dashes are

dhvanipa:
FYI - confirmed e2e objects are produced and synchronizer runs fine

Some docs are outdated and the GUI rendering is broken cuz of the new object structure but I updated those here https://github.com/dobjlabs/digital-objects-network/pull/206

libs/sdk/src/lib.rs
```diff
@@ -655,37 +717,23 @@ impl ActionHandle {
...
-                            .priv_op(Operation::array_contains(
-                                Value::from(in_array.clone()),
-                                idx as i64,
-                                Value::from(pre_dict.clone()),
-                            ))
-                            .unwrap();
-                        array_contains_sts.push(st);
-                    }
-                    if let Some((idx, e)) = meta.out_entry(&varname)
+                    if let Some((idx, e)) = meta.io_entry(&varname)
```
dhvanipa: From Claude (I dont have enough context on this file to verify, also added Rob as reviewer for this reason):

### Bug: the IO merge drops the out-side anchor for Mutate objects

lib.rs:727 replaced the old two per-side ArrayContains pushes with a single lookup through io_entry(). But io_entry returns the first varname match in the chained in_entries → out_entries iterator, and a Mutate object has an entry on both sides. So when a Mutate's out side also needs_wildcard, the rendered predicate demands two anchors (fmt_podlang.rs:426 still emits both), while exec only builds the in-side one. Proof generation then fails on statement count.

Reproduced with a script whose Intro consumes the mutate's final form (the documented reason needs_wildcard exists — Intro args can't be anchored):

```
fn UseFoo(action) {
    var foo = action.mutate("Foo");
    var tick1 = unsafe { foo.tick - 1 };
    action.st_sum(tick1, 1, foo.tick);
    foo.update("tick", tick1);
    var work = action.intro_vdf(3, foo);   // whole-dict pin at final ts
}
```

On main: passes.
On this branch: Wrong statement count for predicate 'UseFoo': expected 6, got 5. The body wants both ArrayContains(io, UseFooIO::in_foo, foo0) and ArrayContains(io, UseFooIO::out_foo, foo); exec emits only the first.

**Fix:** restore two lookups in that loop — in_entry() with the pre-mutation dict and out_entry() with the post dict (out_entry already returns the merged-record slot offset by in_entries.len()). io_entry() then has no callers left and should be deleted;  its first-match-wins semantics is exactly the trap that caused this.

ed255:
> what does `d--` in "kind": "d--" mean? I think d is dictionary but not sure what the dashes are

This is inspired by the read/write/execute flags of files display in Unix.  So `d--` means:
- dictionary: yes
- set: no
- array: no
The reason for using bit flags is because there are cases where a container can be multiple kinds at the same time.  Here's the full context https://github.com/0xPARC/pod2/pull/526

libs/sdk/src/lib.rs
ed255: This is spot-on!
I've fixed the issue and removed the `io_entry` method which is now unused.
https://github.com/dobjlabs/digital-objects-network/pull/205/commits/99201631772ecf65b8102fcf1ece8c602512d3ac

libs/sdk/src/fmt_podlang.rs
robknight: I think this is inaccurate now.

robknight: This refers to in/out, but these are replaced by io.

robknight: This also refers to in/out, not io.

robknight: Outdated comment.

libs/sdk/src/fmt_podlang.rs
```diff
@@ -627,38 +576,32 @@ fn fmt_bridges(loader: &Loader, w: &mut dyn fmt::Write) -> fmt::Result {
             let multi = is_multi_class(&meta.object_refs, &o.class);
             let bridge_name = bridge_predicate_name(&o.class, &meta.name, &o.varname, multi);
 
-            // Bridge predicate signature: state, chain0, chain (public);
+            // Bridge predicate signature: state, state_header, chain0, chain (public);
             // in <ActionIn>, out <ActionOut> private as needed.
```
robknight: Also outdated.

libs/sdk/src/fmt_podlang.rs
robknight: This is the same information as the new RECORD_STATE_HEADER constant - we could derive this from RECORD_STATE_HEADER to ensure there's no drift.

libs/sdk/src/fmt_podlang.rs
```diff
@@ -297,27 +282,10 @@ fn collect_sub_action_calls(action: &ActionContext, loader: &Loader) -> Vec<SubA
...
-                Some(format!("_{}_in_{}", sub_name, idx))
-            } else {
-                None
-            };
-            let sub_out_var = if has_out {
-                Some(format!("_{}_out_{}", sub_name, idx))
-            } else {
-                None
-            };
+            let sub_io_var = format!("_{}_io_{}", sub_name, idx);
```
robknight, line 289: I think the assumption here is that there's always at least one input or output, whereas previously inputs and outputs were optional. Is that true? I think it is, but I want to double-check.

libs/sdk/src/lib.rs
```diff
@@ -1587,10 +1665,20 @@ impl ActionMeta {
...
+    /// Find this Object's `in` entry. Returns its slot in the
+    /// `<Action>In` record and the entry shape.
+    pub(crate) fn in_entry(&self, varname: &str) -> Option<(usize, &EntryShape)> {
+        self.in_entries
+            .iter()
+            .enumerate()
+            .find(|(_, e)| e.varname == varname)
+    }
+
     pub(crate) fn out_entry(&self, varname: &str) -> Option<(usize, &EntryShape)> {
```
robknight, line 1681: This returns an index into the combined in/out vectors, right? Whereas in_entry returns an index into the `in_entries` vector. This works because we know that these two vectors will be combined, with `in_entries` before `out_entries`? A comment to explain this might be helpful.

libs/sdk/src/tests.rs
```diff
@@ -562,3 +556,76 @@ fn Bad(action) {{
...
+    let [ticker0] = res.objs();
+    apply_tx(&mut state, &ticker0_tx);
+
+    println!("exe Tick");
+    state.next_block(4000);
+    let executor = module.executor(true, grounding_witness(&state, &[ticker0.obj.commitment()]));
+    let res = executor.action("Tick", vec![ticker0]).unwrap();
+    let ticker1_tx = res.tx.clone();
+    let [_ticker1] = res.objs();
+    apply_tx(&mut state, &ticker1_tx);
```
robknight, line 630: There are no asserts in this test. Are we just testing for panics?

libs/sdk/src/lib.rs
```diff
@@ -452,13 +479,46 @@ impl ActionContext {
...
 impl ActionHandle {
     //
     // Internal methods
     //
     fn new(name: String, exe_ctx: Option<Rc<RefCell<ExeContext>>>) -> Self {
         Self(Rc::new(RefCell::new(ActionContext::new(name, exe_ctx))))
     }
+    fn state_header(&self) -> ArgHandle {
+        let mut ctx = self.0.borrow_mut();
+        ctx.add_var("state_header".to_string()).unwrap_or(());
```
robknight, line 507: What are the circumstances where this could fail?

libs/sdk/src/fmt_podlang.rs
```diff
@@ -627,38 +576,32 @@ fn fmt_bridges(loader: &Loader, w: &mut dyn fmt::Write) -> fmt::Result {
...
             // in <ActionIn>, out <ActionOut> private as needed.
-            write!(w, "{bridge_name}(state, chain0, chain")?;
+            write!(w, "{bridge_name}(state, state_header, chain0, chain")?;
             let mut priv_parts: Vec<String> = Vec::new();
-            if !meta.in_entries.is_empty() {
-                priv_parts.push(format!("in {}", schema_name(&meta.name, Side::In)));
-            }
-            if !meta.out_entries.is_empty() {
-                priv_parts.push(format!("out {}", schema_name(&meta.name, Side::Out)));
+            if !meta.in_entries.is_empty() || !meta.out_entries.is_empty() {
```
robknight: When making subaction calls, we assume that there's always an `io`, but here we assume that there might not be one?

libs/txlib/src/lib.rs
```diff
@@ -51,7 +51,7 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
 /// root hash and to verify synchronizer-supplied membership proofs. Full
 /// containers are not carried -- callers prove each input's liveness with a
 /// per-object Merkle proof packaged in a [`GroundingWitness`].
-#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
```
robknight: When is a default StateHeader useful?

robknight (changes requested):
Seems good overall, just some questions about how `io` is handled and some outdated comments which need tidying up.

libs/sdk/src/fmt_podlang.rs
ed255: Yes.  I propose we assume there's always an `io` wildcard with at least one input or output.  I'll review the code because I think it doesn't always make this assumption.

libs/sdk/src/lib.rs
ed255: Your explanation is right.  My alternative explanation is that these methods return the position in the io record.  I'll add a comment with the explanation.

libs/sdk/src/tests.rs
ed255: The intention here is to run the action and make sure it succeeds and the statements are built satisfied.  So yeah, we're testing for panics because we're unrwapping the result of `executor.action`.  Do you think some asserts would be useful here?

libs/sdk/src/lib.rs
ed255: This is idempotent.  If the var doesn't exist it will be added.  If it exists, `add_var` will return `Err(format!("var {var} already exists").into())` and `unwrap_or` will discard the error.

libs/sdk/src/fmt_podlang.rs
ed255: You're right, this was an inconsistency.  I've removed this conditional check, now it's unconditional.

libs/txlib/src/lib.rs
ed255: I added this for tests, but I realized that it's not used, so I'll remove it.

libs/sdk/src/fmt_podlang.rs
ed255: done!

ed255:
I've merged https://github.com/dobjlabs/digital-objects-network/pull/206 into this PR's branch, which fixed most (if not all) of the documentation issues @robknight mentioned.

libs/sdk/src/tests.rs
robknight: No, I just wanted to check the rationale.

robknight (approved):
lgtm, the merge addresses the issues I raised.

PR #206, opened 2026-07-30
Update docs and fix gui

dhvanipa:


interfaces/gui/src/shared/pod2utils.ts
```diff
@@ -61,6 +65,68 @@ function normalizePod2ContainerInner(value: unknown): unknown {
...
+function normalizePod2Container(kind: string, kvs: unknown[]): unknown {
+  const entryPair = (entry: unknown): [unknown, unknown] | null => {
+    if (!Array.isArray(entry) || entry.length < 1 || entry.length > 2) {
+      return null;
+    }
+    return [entry[0], entry[entry.length - 1]];
+  };
+  const pairs = kvs.map(entryPair);
+
+  if (kind[0] === "d") {
```
ed255, line 99: I'm not 100% sure this handles all the edge cases.

Consider this dictionary: `{"a": 0, "b": "b"}`.  The kvs will be `[["a", 0], ["b"]]`, that is, the second entry is a single element vector, but the `map` takes `[key, entry]`.  Will that throw an exception?

Something similar can happen with an array.

ed255 (approved):
Overall looks good to me!  I've added some notes on the deserialization of containers because I'm not sure they work with the edge cases described, please take a look.

PR #210, opened 2026-08-04
sdk: intro pods error handling

### LINKED ISSUE #209: LtEqU256 constraint failure panics and terminates dobjd
When a real LtEqU256 comparison fails because lhs > rhs, ActionHandle::intro_lt_eq_u256 unwraps the resulting error and panics. dobjd terminates instead of reporting a failed action.

Create an action that compares a nonzero object value against zero:

```rhai
fn RejectLargeCommitment(action) {
    var output = action.output("Output");
    let zero = action.top_limb_u256(0);
    action.intro_lt_eq_u256(output, zero);
}
```

Run it with the real prover (`mock = false`). The comparison returns `lhs > rhs in LtEqU256`, after which the SDK panics at this unwrap:

```rust
LtEqU256Pod::new_boxed(...).unwrap();
```

A secondary Rhai argument-cleanup panic may appear during unwinding.

Tested with `v0.1.0-rc.43` on `x86_64-pc-windows-msvc`.

ed255:
Resolve #209

### DIFF
--- a/libs/sdk/src/error.rs
+++ b/libs/sdk/src/error.rs
@@ -1,3 +1,5 @@
+use std::rc::Rc;
+
 use rhai::{EvalAltResult, Position};
 
 #[derive(Debug, thiserror::Error)]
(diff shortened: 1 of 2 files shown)


robknight (approved):
lgtm!

PR #211, opened 2026-08-04
Support mutating multiple objects and reading from mutated objects

robknight:
This adds SDK support for actions which mutate multiple objects, and for reading from mutated objects in sub-actions.

This does _not_ add support for reading from multiple objects, which seems likely to be needed in future but is not a current requirement (as far as we know).

### DIFF
--- a/libs/sdk/src/lib.rs
+++ b/libs/sdk/src/lib.rs
@@ -153,11 +153,17 @@ enum Inst {
     /// statement is consumed here in the parent's post-Rhai walk.
     SubAction {
         action: String,
-        /// Aliases the sub-action's first producing object Ref so
-        /// parent scripts can bind it via `var foo = subaction(...)`.
+        /// Aliases the Ref of the sub-action's first produced object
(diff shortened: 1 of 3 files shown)


libs/sdk/src/lib.rs
```diff
@@ -1172,43 +1218,64 @@ impl ActionHandle {
...
+                let entry = sub_io_array
+                    .get(entry_idx)
+                    .expect("array op")
+                    .expect("sub io record has an entry per produced object");
+                // An Output's out entry is the post-identity dict, but
+                // the sub left its Ref at the pre-identity form; rebind
+                // so parent reads see the recorded form.
+                if io == ObjectIO::Output {
+                    obj.borrow_mut().set_value(entry.clone());
+                }
```
ed255, line 1255: This looks very subtle and tricky, as in easy to forget and hard to debug if this was not here.

ed255 (approved):
LGTM!

PR #214, opened 2026-08-14
Local dev environment using anvil

### LINKED ISSUE #26: add ability to run with local chain

robknight:
This adds a new `just dev-local` command, which sets up a local dev environment using [anvil](https://www.getfoundry.sh/anvil) instead of remote services for accessing blob space. There is also a shim which adds support for some simple APIs that anvil doesn't support natively.

`just dev-local` sets up new databases for the synchronizer and relayer, so you can switch back and forth between local and remote dev without overwriting the synchronizer/relayer data.

Local block time is 2 seconds, which speeds up local testing a bit. You now don't have to wait very long for transactions to finalize. This is also pretty useful for doing live demos.


devtools/beacon-shim/src/main.rs
```diff
@@ -0,0 +1,341 @@
...
+/// `eth-clients` sends repeated query pairs, which would reach anvil as the
+/// last pair alone and silently return the wrong blob set.
+async fn fetch_blobs(state: &AppState, block_id: &str, versioned_hashes: &[B256]) -> Result<Bytes> {
+    let mut url = format!("{}/eth/v1/beacon/blobs/{}", state.anvil_url, block_id);
+    if !versioned_hashes.is_empty() {
+        let joined: Vec<String> = versioned_hashes.iter().map(|h| h.to_string()).collect();
+        url.push_str(&format!("?versioned_hashes={}", joined.join(",")));
+    }
+    let response = state.http.get(&url).send().await?;
+    if response.status() == StatusCode::NOT_FOUND {
```
dhvanipa, line 330: An upstream 404 should mean the block doesnt exist right? This converts it into an empty block which I think is incorrect

For blocks with no blobs anvil should already be returning 200

this should return a valid error instead

dhvanipa:
> so you can switch back and forth between local and remote dev without overwriting the synchronizer/relayer data.

they share the same ~/.dobj folder so technically the objects would not work across both anyways right? is there a reason to not just use the same db for simplicity? or is there a reason as a dev you want both?

devtools/beacon-shim/src/main.rs
```diff
@@ -0,0 +1,341 @@
...
+    }
+}
+
+impl IntoResponse for ApiError {
+    fn into_response(self) -> Response {
+        match self {
+            Self::NotFound => (StatusCode::NOT_FOUND, "not found").into_response(),
+            Self::Internal(err) => {
+                warn!(?err, "Beacon shim request failed");
+                (StatusCode::INTERNAL_SERVER_ERROR, err.to_string()).into_response()
```
dhvanipa: This returns `err.to_string()` as a bare body, but every eth-clients caller parses non-404 responses as `{code, message}` JSON, so this will lose the error

See https://github.com/dobjlabs/digital-objects-network/blob/main/libs/eth-clients/src/common.rs#L219, https://github.com/dobjlabs/digital-objects-network/blob/main/libs/eth-clients/src/common.rs#L37

devtools/beacon-shim/src/main.rs
```diff
@@ -0,0 +1,341 @@
...
+    #[serde(default)]
+    versioned_hashes: Vec<B256>,
+}
+
+async fn get_blobs(
+    Path(block_id): Path<String>,
+    Query(query): Query<BlobsQuery>,
+    State(state): State<AppState>,
+) -> Result<Response, ApiError> {
+    let body = fetch_blobs(&state, &block_id, &query.versioned_hashes).await?;
```
dhvanipa: A comment From Claude:

The shim forwards `head` and `finalized` directly to Anvil, but Anvil's blob endpoint only accepts a concrete block identifier. In the smoke test `/eth/v1/beacon/blobs/head` returned 500 because Anvil returned 400. Resolve the requested `BlockId` first and forward the resulting block hash, as the header and block handlers already do

--

I see `resolve_block` which makes me think this comment is not true, but then I wonder why Claude said it so just sharing it

dhvanipa (approved):
besides comments lgtm! Fixes https://github.com/dobjlabs/digital-objects-network/issues/26, and I'm glad the fix is simpler than a running a whole execution node locally via docker

robknight:
> > so you can switch back and forth between local and remote dev without overwriting the synchronizer/relayer data.
> 
> they share the same ~/.dobj folder so technically the objects would not work across both anyways right? is there a reason to not just use the same db for simplicity? or is there a reason as a dev you want both?

I've made it so that they're entirely separate, including the dobj folder. My thinking here is that if you have a real dobj folder and real synchronizer, you don't want them to pick up state from the local testing network. However, I'd be happy to change it in the opposite direction if you prefer.

dhvanipa:
> > > so you can switch back and forth between local and remote dev without overwriting the synchronizer/relayer data.
> > 
> > 
> > they share the same ~/.dobj folder so technically the objects would not work across both anyways right? is there a reason to not just use the same db for simplicity? or is there a reason as a dev you want both?
> 
> I've made it so that they're entirely separate, including the dobj folder. My thinking here is that if you have a real dobj folder and real synchronizer, you don't want them to pick up state from the local testing network. However, I'd be happy to change it in the opposite direction if you prefer.

Entirely separate works too! lgtm

PR #220, opened 2026-08-18
Support anchored key arguments to intro predicates

### LINKED ISSUE #215: SDK compile/execution mismatch for anchored arguments in intro statements
Microverse passes an object dictionary field directly to `intro_lt_eq_u256` in its [range-extraction helper](https://github.com/silentwerker/microverse/blob/af73ff464d80ed0f2ef9b3a6c379c976e4810854/plugin.rhai#L782-L784), used by [`ExtractRockyPlanetIgneousRock`](https://github.com/silentwerker/microverse/blob/af73ff464d80ed0f2ef9b3a6c379c976e4810854/plugin.rhai#L3001).

The SDK compiler [renders dictionary-field references as anchored entries](https://github.com/dobjlabs/digital-objects-network/blob/02bee560249616d101ec68490696d9b34d6b4121/libs/sdk/src/fmt_podlang.rs#L175-L190), but during execution [`intro_lt_eq_u256` constructs the intro pod from concrete values](https://github.com/dobjlabs/digital-objects-network/blob/02bee560249616d101ec68490696d9b34d6b4121/libs/sdk/src/lib.rs#L1372-L1394). The literal intro statement is then [attached with `add_pod(...).unwrap()`](https://github.com/dobjlabs/digital-objects-network/blob/02bee560249616d101ec68490696d9b34d6b4121/libs/sdk/src/lib.rs#L1408-L1412), producing:

```text
LtEqU256(Raw(lower), Raw(actual))
doesn't match
LtEqU256(Raw(lower), body0["source_signal_identifier"])
```

[POD2 PR #519](https://github.com/0xPARC/pod2/pull/519) explicitly supports anchored keys in intro statements via `ReplaceValueWithEntry`, and the current POD2 revision includes that change.

This appears distinct from [issue #209](https://github.com/dobjlabs/digital-objects-network/issues/209) and [PR #210](https://github.com/dobjlabs/digital-objects-network/pull/210), which handle errors returned while constructing an intro pod. Here construction succeeds and the failure occurs while matching/attaching its statement.

robknight:
Fixes #215 

Since all arguments can now be anchored keys, the `blocks_initials` mechanism is no longer needed. The `initials` record is a record of all initial object states (before insertion). This could previously be "blocked" by the existence of an object whose initial state needs to be passed to an intro statement, because intro statements could not take anchored key arguments from the `initials` record. Instead, separate wildcard arguments were required. Now they are not required, intro statements can use `initials.key` directly, at the cost of an extra `ReplaceValueWithEntry` operation.

### DIFF
--- a/examples/craft-basics/manifest.toml
+++ b/examples/craft-basics/manifest.toml
@@ -2,7 +2,7 @@
 name = "craft-basics"
 version = "0.1.0"
 # Rewritten by `cargo run -p pexe -- build examples/craft-basics`.
-module_hash = "57631b51fb9a921588d391211f94c0bd8f777aff0a16755bc2dfefb52d6ff5b0"
+module_hash = "d7edcb9150d12af76a54fbbac7b00b8bb24fab61bc1395b65da47547ad5d1b42"
 
 [[classes]]
 name = "Log"
--- a/examples/craft-rocket/manifest.toml
+++ b/examples/craft-rocket/manifest.toml
@@ -2,7 +2,7 @@
 name = "craft-rocket"
 version = "0.1.0"
 # Rewritten by `cargo run -p pexe -- build examples/craft-rocket`.
-module_hash = "f838acbb5b8a196174ec1adb15b2dc19f60faa5ef76ab0ab60b6ee010f0f964f"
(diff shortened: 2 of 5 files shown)


ed255 (approved):
LGTM!

PR #223, opened 2026-08-21
SDK: host methods for most native statements

robknight:
This PR adds Rhai host methods for most native POD2 statements, excluding `SignedBy` and `PublicKey` (it's not clear that we want to support these anyway).

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -162,12 +162,12 @@ The driver does not cache compiled modules between calls — every `Driver::exec
 
 - `var <ident> = <expr>` — declare a wildcard. Operations on it generate constraining statements.
 - `let <ident> = <expr>` — plain Rhai, literal known at both phases. No statements emitted.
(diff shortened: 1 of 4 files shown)


dhvanipa (approved):
lgtm! would we not want signedby and publickey to be able to create objects that only an admin could mint for eg, or a guild could control? I feel like it will be useful

robknight:
> lgtm! would we not want signedby and publickey to be able to create objects that only an admin could mint for eg, or a guild could control? I feel like it will be useful

I suppose so, yes. Right now we don't have any tools for creating keypairs, or for key management (e.g. storing your keys alongside your objects). I think it needs some thought.

PR #227, opened 2026-08-24
Initial cross-module actions

### LINKED ISSUE #113: ability to refer an action from one pexe in another

robknight:
Closes #113 

Cross-module actions. Pexe plugins can now declare dependencies on other pexe plugins.

A dependency declaration has two parts: a module hash, and an alias. The alias is the name which is used locally to refer to the dependency - it is a purely local alias which is used in the Rhai script to refer to the dependency's classes/actions by name, and in the generated Podlang. Aliases must be locally unique (a plugin cannot use the same alias for two different imports) but does not need to be globally unique (two different plugins might use the same alias to refer to different dependencies, including different "versions" of the "same" plugin).

In the future we might want to introduce a package registry, which would give us globally unique names, and semantic version numbers for plugins. However, a content-based approach works for now.


examples/craft-totem/manifest.toml
```diff
@@ -0,0 +1,22 @@
...
+module_hash = "d854e4fb383457f9f06a61db92975111a0e7384aed3edba58ac2899808d978ee"
+
+# craft-totem calls craft-basics actions via subaction("craft-basics::...").
+# The pin is stamped by `pexe build`, which resolves the import among
+# already-built archives (build craft-basics first).
+[[imports]]
+name = "craft-basics"
+module_hash = "d7edcb9150d12af76a54fbbac7b00b8bb24fab61bc1395b65da47547ad5d1b42"
+
+[[classes]]
```
dhvanipa, line 14: does this PR also add the ability to create pexe's without declaring a new class?

examples/craft-totem/manifest.toml
```diff
@@ -0,0 +1,22 @@
...
+name = "craft-totem"
+version = "0.1.0"
+# Rewritten by `cargo run -p pexe -- build examples/craft-totem`.
+module_hash = "d854e4fb383457f9f06a61db92975111a0e7384aed3edba58ac2899808d978ee"
+
+# craft-totem calls craft-basics actions via subaction("craft-basics::...").
+# The pin is stamped by `pexe build`, which resolves the import among
+# already-built archives (build craft-basics first).
+[[imports]]
+name = "craft-basics"
```
dhvanipa, line 11: without a path to the code how is the code resolved? does it loop in the directory? curious why you went for this approach vs making the user specify the path

dhvanipa: ahh I see there is `--deps` so this is kind of like specifying a path but not at the pexe level

libs/driver/src/pexe_catalog.rs
```diff
@@ -341,17 +380,25 @@ impl ActionCatalog for PexeCatalog {
...
-        return Ok(Vec::new());
-    }
-    let mut entries: Vec<PathBuf> = std::fs::read_dir(actions_dir)
-        .with_context(|| format!("failed to read {}", actions_dir.display()))?
-        .filter_map(|e| e.ok().map(|e| e.path()))
-        .filter(|p| p.extension().and_then(|s| s.to_str()) == Some(pexe::PEXE_EXTENSION))
-        .collect();
-    entries.sort();
+thread_local! {
+    /// Per-thread module cache, keyed by validated manifest hash.
```
dhvanipa, line 384: Does this get refreshed on new installs? ie when `reload_catalog` is called in driver

dhvanipa (approved):

examples/craft-totem/manifest.toml
robknight: That concept doesn't make any sense to me. Declaring new classes is what pexes are for!

dhvanipa: in PAC, agents negotiate a pexe on the fly which is "throwaway", its just meant to represent a swap of 2 objects

so all it has is

fn Swap(){
 rekeyA()
 rekeyB()
 }

examples/craft-totem/manifest.toml
ed255, line 11: This `name` is the alias right?  I would suggest naming this field `alias`.  I see this value stored in a variable named `alias` in the implementation.  Otherwise one may think that the `name` must match the original module name.

libs/driver/src/pexe_catalog.rs
```diff
@@ -5,23 +5,20 @@
...
-//! in its own plugin.
+//! Classes and actions are keyed by [`QualifiedName`] (`<plugin>::<name>`), so
+//! plugins may reuse bare names. Cross-plugin operations use manifest imports
+//! and qualified `subaction` calls. Imports load in dependency order and are
+//! validated against their pinned module hashes.
 //!
-//! The compiled [`sdk::SdkModule`] is not kept — it holds a `Rc<Engine>` and is
-//! therefore `!Send`. `execute_action` re-loads the script from its stored bytes
-//! on demand, matching the per-call pattern used before.
+//! [`sdk::SdkModule`] is not `Send`, so execution uses a thread-local cache of
```
ed255, line 13: Rhai has a feature to make its types thread-safe (`Send`).  If we need it we could enable it.  Doing that would require replacing a bunch of `Rc`s for `Arc`s (I'm not sure about the performance impact of that change).
That would probably allow making the `SdkModule` `Send`.

libs/driver/src/pexe_catalog.rs
```diff
@@ -284,6 +304,33 @@ impl PexeCatalog {
...
+    fn compiled_module(&self, plugin: &Plugin) -> Result<Rc<sdk::SdkModule>> {
+        let plugin_name = plugin.manifest.plugin.name.as_str();
+        let module_hash = plugin.manifest.plugin.module_hash;
+        let cached = COMPILED_MODULES.with(|cache| cache.borrow().get(&module_hash).cloned());
+        if let Some(module) = cached {
+            return Ok(module);
+        }
+        let sdk = Sdk::default();
+        let mut resolver = plugin_resolver(&sdk, &self.plugins);
+        let module = resolver
```
ed255, line 319: This function caches the requested module, but it doesn't cache the imported modules right?
So if I have these two module graphs:
- `A <- B` (A imports B)
- `C <- B` (C imports B)

And I call `compiled_module(A)`, B and A are compiled and then `COMPILED_MODULES = {A}`
Then I call `compiled_module(C)`, B and C are compiled (so B is compiled again) and then `COMPILED_MODULES = {A, C}`.

I wonder if a small change could allow us to store `B` in the thread-local cache, and have the plugin resolver also use it, so that in the scenario I described, `B` is not compiled twice.

libs/pexe/README.md
```diff
@@ -57,18 +54,28 @@ cargo run -p pexe --release -- build --install \
 # Fail if manifest module_hash doesn't match compiled hash
 # (by default the source manifest gets rewritten to match)
 cargo run -p pexe --release -- build --check examples/*
+
+# Resolve [[imports]] against archives in an extra directory, searched
+# before target/pexe and the install dir
+cargo run -p pexe --release -- build --deps /path/to/archives examples/craft-totem
```
ed255, line 60: Can we add multiple `--deps` flags in the cli call?
I believe the answer is yes, as `deps` is a vector in the cli struct.

libs/pexe/src/inspect.rs
```diff
@@ -495,18 +523,10 @@ fn print_dep_graph(plan: &sdk::PlanData, aliases: &HashMap<Hash, String>) {
...
+/// Includes transitive imports and omits the local module.
 fn build_alias_map(module: &SdkModule) -> HashMap<Hash, String> {
-    let mut map = HashMap::new();
-    for dep in module.dependencies() {
-        if let Dependency::Module { name, hash } = dep {
-            map.insert(*hash, name.clone());
-        }
-    }
-    map
+    module.module_aliases()
```
ed255, line 529: Maybe `build_alias_map` can be deleted?  Writing `module.module_aliases()` is similarly short.

libs/sdk/src/fmt_podlang.rs
```diff
@@ -19,8 +19,14 @@ use txlib::RECORD_STATE_HEADER_PODLANG;
 
 fn fmt_dependency(dep: &Dependency, w: &mut dyn fmt::Write) -> fmt::Result {
     match dep {
-        Dependency::Module { name, hash } => {
-            writeln!(w, "use module {:#} as {name}", hash)?;
+        Dependency::Module { alias, hash } => {
+            // An alias may contain `-`; a podlang identifier may not.
```
ed255, line 23: Is allowing `-` in aliases inspired by how cargo works?
I would suggest not allowing `-` in aliases to make things simpler.

libs/pexe/src/lib.rs
```diff
@@ -127,37 +132,243 @@ fn read_entry<R: Read + std::io::Seek>(zip: &mut ZipArchive<R>, name: &str) -> R
...
+            // A bad optional search directory does not block other candidates.
+            log::warn!("skipping dependency search dir: {err}");
+            Vec::new()
+        })
+    });
+    for path in dep_paths {
+        match read_pexe_file(&path).and_then(|bytes| unpack(&bytes)) {
+            Ok((manifest, script)) => sources.push((manifest, script)),
+            Err(err) => {
+                log::warn!("skipping unreadable pexe {}: {err}", path.display());
```
ed255, line 204: I feel like it would be better to error here, so the user is forced to check what's happening and remove or fix the corrupted pexe.
On the other hand, this is probably running from the daemon and getting this error to the user may be not be easy?

libs/pexe/src/lib.rs
```diff
@@ -127,37 +132,243 @@ fn read_entry<R: Read + std::io::Seek>(zip: &mut ZipArchive<R>, name: &str) -> R
...
+    let tables: Vec<&mut dyn toml_edit::TableLike> = match imports {
+        toml_edit::Item::ArrayOfTables(tables) => tables
+            .iter_mut()
+            .map(|table| table as &mut dyn toml_edit::TableLike)
+            .collect(),
+        toml_edit::Item::Value(toml_edit::Value::Array(values)) => values
+            .iter_mut()
+            .filter_map(|value| value.as_inline_table_mut())
+            .map(|table| table as &mut dyn toml_edit::TableLike)
+            .collect(),
```
ed255, line 360: What is this form?  Why do we support 2 forms?
I think that I've only seen the `[[imports]]` form in the docs and comments, which is the `ArrayOfTables`.

libs/sdk/src/lib.rs
```diff
@@ -46,8 +45,92 @@ type RuntimeResult<T> = Result<T, Box<EvalAltResult>>;
...
+/// Aliases qualify `subaction` calls and generated podlang. Module identity
+/// comes from the module's batch ID, not the alias.
+#[derive(Clone)]
+pub struct ModuleImport {
+    pub alias: String,
+    pub module: Rc<SdkModule>,
+}
+
+/// Convert an import alias to a podlang identifier.
+fn podlang_alias(alias: &str) -> String {
```
ed255, line 71: See my comment about not allowing `-` in aliases.

libs/sdk/src/lib.rs
```diff
@@ -1858,18 +2000,79 @@ fn try_value_from_dynamic(v: Dynamic) -> RuntimeResult<Value> {
...
+            defining: self.defining.clone().or_else(|| defining.cloned()),
+            ..self.clone()
+        }
+    }
+}
+
+impl fmt::Display for ActionObjectRef {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match &self.defining {
+            Some(defining) => write!(f, "{defining:?}::{}", self.class),
```
ed255, line 2070: `DefiningModule` contains `SdkModule`.  Is it a good idea to do a Debug display here?  Won't it be too crowded with data?

libs/sdk/src/lib.rs
```diff
@@ -2045,19 +2252,28 @@ impl ActionMeta {
...
+                    });
+                    meta.total_inputs.extend(
+                        sub.total_inputs
+                            .iter()
+                            .map(|object_ref| object_ref.qualified_by(defining.as_ref())),
+                    );
+                    meta.total_outputs.extend(
+                        sub.total_outputs
+                            .iter()
+                            .map(|object_ref| object_ref.qualified_by(defining.as_ref())),
```
ed255, line 2275: I just realized that `ActionObjectRef` has `defining: Option<DefiningModule>`, and because it's optional we need to handle both cases in several places.
What if the action when called from the module that defines it also contains `DefiningModule`?  Internally it has an `Rc<SdkModule>` so having copies is cheap.  And then we don't need to handle 2 different cases.
To be consistent, we could have a special alias to refer to the local module, like `root`?  I've seen the mention of `root` as the module name of the module where actions are defined.

libs/sdk/src/lib.rs
```diff
@@ -2275,22 +2492,90 @@ impl Loader {
...
+                    "import alias {:?} collides with the reserved tx module alias",
+                    import.alias
+                ));
+            }
+            if PODLANG_RESERVED_WORDS.contains(&alias.as_str()) {
+                return Err(anyhow!(
+                    "import alias {:?} maps to {alias:?}, which podlang reserves",
+                    import.alias
+                ));
+            }
```
ed255, line 2520: it would be nice to move all this code to a `validate_alias` function

libs/sdk/src/lib.rs
```diff
@@ -2400,6 +2684,43 @@ impl SdkModule {
...
+    /// names. Direct-import aliases override aliases found transitively. The
+    /// current module is omitted so its predicates render unqualified.
+    pub fn module_aliases(&self) -> HashMap<Hash, String> {
+        let mut aliases = HashMap::new();
+        self.collect_module_aliases(&mut aliases);
+        aliases
+    }
+    fn collect_module_aliases(&self, aliases: &mut HashMap<Hash, String>) {
+        // Traverse children first so direct aliases take precedence.
+        for import in &self.imports {
```
ed255, line 2697: Question for myself:
Imports are declared in the manifest.
Where are dependencies declared?

libs/sdk/src/lib.rs
```diff
@@ -2934,18 +3373,53 @@ impl Sdk {
...
     pub fn load_module_from_src_manifest(
         &self,
         src: &str,
         manifest: &Manifest,
+        imports: &[ModuleImport],
     ) -> Result<Rc<SdkModule>, SdkError> {
+        for declared in &manifest.imports {
+            let import = imports
+                .iter()
+                .find(|import| import.alias == declared.name)
```
ed255, line 3391: From an earlier comment, I think `declared.alias` would make this more clear.

ed255 (approved):
Overall LGTM!  I've left a few comments with various suggestions

PR #228, opened 2026-08-24
Pexe inspect plan fixes

### LINKED ISSUE #222: `pexe inspect plan` panics after synthesizing invalid fixtures for valid actions
## Summary
`pexe inspect plan` panics before producing a plan for valid actions with repeated input classes or deeply split producer schemas.

## Reproduction

Using the [Microverse PEXE](https://github.com/silentwerker/microverse/blob/af73ff464d80ed0f2ef9b3a6c379c976e4810854/microverse.pexe):

```console
pexe inspect plan microverse.pexe --action BuildShipMedium --show header,summary,totals
pexe inspect plan microverse.pexe --action BuildShipLarge --show header,summary,totals
```

Observed:

- `BuildShipMedium`: `Frontend(Inner: invalid arguments to SumFromEntries operation)`
- `BuildShipLarge`: SDK panic `key exists` because required `ShipyardPermit` fields are absent

Both actions plan and mock-execute successfully when supplied correctly shaped, slot-specific fixtures.

## Cause

[`prepare_run`](https://github.com/dobjlabs/digital-objects-network/blob/02bee560249616d101ec68490696d9b34d6b4121/libs/pexe/src/inspect.rs#L294-L305) sends only input class names to `mint_classes`. [`mint_classes`](https://github.com/dobjlabs/digital-objects-network/blob/02bee560249616d101ec68490696d9b34d6b4121/libs/pexe/src/fixtures.rs#L36-L80) therefore reuses class-wide defaults for repeated `Resource` slots instead of honoring each action slot's literal constraints.

Separately, [`derive_class_signature` / `inline_action`](https://github.com/dobjlabs/digital-objects-network/blob/02bee560249616d101ec68490696d9b34d6b4121/libs/pexe/src/inspect.rs#L1158-L1241) expands only one `BatchSelf` level, so fields introduced earlier in a split producer chain can be omitted from synthetic objects.

## Expected

`inspect plan` should synthesize complete, action-slot-aware inputs, or return an actionable unsupported-fixture error rather than panic. Regression coverage should include two same-class inputs with different exact literals and a multi-level split producer whose consumer reads early and late fields.

Related: [PR #129](https://github.com/dobjlabs/digital-objects-network/pull/129) fixed missing object identity in inspect fixtures, but not these per-slot constraint and recursive schema problems.

robknight:
Closes #222 

Fixes `pexe inspect` so that instead of using the low-level Podlang predicates, it uses the high-level Rhai AST to infer information about object/action structure. This is used for:
- `pexe inspect plan` - explains how a real transaction proof for a given action would be built, e.g. which statements it uses and how many PODs
- `pexe inspect prove` - generates a real proof from mock data
- `pexe inspect classes` - lists classes and their fields and constraints (where possible to infer)

This previously worked for the limited examples we had, but worked much less well for the larger `microverse` example, with panics and excessive memory usage. Switching to the more reliable source of information about the actions/classes solves the issue.


libs/sdk/src/requirements.rs
```diff
@@ -0,0 +1,618 @@
...
+        self.vdf |= other.vdf;
+        self.proof_of_work |= other.proof_of_work;
+    }
+}
+
+/// Constraints and type requirements imposed on a field by an action.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct FieldFacts {
+    /// Literal values this field is constrained to equal.
+    pub pinned: BTreeSet<Pin>,
```
ed255, line 56: why is this a set and not a single value?

libs/sdk/src/requirements.rs
```diff
@@ -0,0 +1,618 @@
...
+use std::fmt;
+use std::rc::Rc;
+
+use pod2::middleware::{NativePredicate, Value};
+
+use crate::{ActionContext, Inst, Intro, Ref, Var, VarOrValue, arg_is_int};
+
+/// Literal value constraint for a field.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub enum Pin {
```
ed255, line 17: Why declare `Pin` instead of using `Value`?

libs/sdk/src/requirements.rs
```diff
@@ -0,0 +1,618 @@
...
+    }
+}
+
+/// Constraints and type requirements imposed on a field by an action.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct FieldFacts {
+    /// Literal values this field is constrained to equal.
+    pub pinned: BTreeSet<Pin>,
+    /// Minimum allowed integer value.
+    pub min: Option<i64>,
```
ed255, line 58: There's min but not `max`.

I would like to have a discussion about the idea behind this representation, and what's the future plan for this `requirements` feature.

Technically we could just show the list of predicates and that would show the requirements.  But I guess here you're aiming for something more readable?
Is the goal to capture as much as you can but not everything here?
For example, if we have a `lt` statement you can fill in the `min` value.  But if we have something like`a = 2*b`, would you store that in a `FieldFact` for `a`?  Or the goal is never to be exhaustive?

ed255 (commented):
I have some questions about the idea behind the feature that gathers requirements.

libs/sdk/src/requirements.rs
ed255: Here's a summary of what I discussed in yesterdays call:
In general I think the `inspect` functionality is nice and useful, so I'm in favor of keeping it.
One of the features is to report requirements around actions, which is mostly implemented in the `requirements.rs` file now (previously the same feature set was implemented in another file).  The way it is implemented seems a bit complicated to me: until now he had a language to express requirements, which is the statement templates (sets of statements that have wildcards + literals as arguments).  This file implements a new shape to define requirements and I find it complex because
- It doesn't use the "statement template" language, and instead it introduces a new custom language to express observed requirements or constraints.
- It needs to be extended to cover more information: for example now there's `min` but no `max`
- I believe it can never be complete, because it's too simple.  It can express relations to constants but not relations to variables.

For this reason I would suggest removing this "custom language" for requirements.  This includes `FieldFacts`, `ObjectIdentity`, `FieldWrites` and other types as well as associated functions to collect this information.  Instead I suggest following a simpler approach that would convey the same information albeit in a less easy-digestible way: build a map from object fields to the list of statement templates that constrain them and display that as facts.

PR #235, opened 2026-08-26
SDK: add literal containers and variable-key lookups

robknight:
This adds first-class container literals and variable-key lookups to the SDK, making it practical to represent lookup tables directly in Rhai scripts.

Rhai arrays and maps now convert recursively to pod2 Arrays and Dictionaries, while the new free-standing set_of function constructs Sets. Scripts can query these containers with `array_get`, `dict_get`, or `obj.get`, including when the key is a variable.

## Key changes

- Add recursive Array, Dictionary, and Set literal conversion.
- Add variable-key lookups backed by containment statements.
- Allow SDK-generated variables to be used inline.
- Allow var to name literals without creating wildcards.
- Snapshot statement and write arguments before later object mutations.
- Validate container kinds, lookup key types, and literal lookups during Load where possible.
- Report invalid field access as script errors instead of panicking.
- Render container literals deterministically, with commitment fallbacks for sparse arrays.
- Update the SDK documentation and regression coverage.
- Add the `nanoverse` example plugin to demonstrate lookup tables, witness arithmetic, state-header access, subactions, and related SDK features.

These changes let authenticated object data select rows from literal tables without requiring a separate action for every row. Player-selected choices can remain separate actions, while data-selected behavior can be expressed compactly through constrained lookups.


libs/sdk/src/fmt_podlang.rs
```diff
@@ -184,11 +185,79 @@ struct ArgFmt<'a> {
...
 }
 
+/// Formats a value as a Podlang literal.
+///
+/// Scalars already use valid Podlang syntax, but `Value` formats containers
+/// using debug syntax. If a container can be interpreted as multiple kinds,
+/// prefer set, then dictionary, then array. Each representation has the same
+/// raw value, which is what statement arguments use.
+///
+/// Podlang cannot represent sparse arrays, so they are rendered as their
```
ed255, line 195: Add a note about https://github.com/0xPARC/pod2/pull/541 here

libs/sdk/src/fmt_podlang.rs
```diff
@@ -184,11 +185,79 @@ struct ArgFmt<'a> {
...
+            slots.sort_by_key(|(index, _)| *index);
+            // Podlang array literals cannot represent gaps between indexes.
+            if slots.iter().enumerate().all(|(i, (index, _))| i == *index) {
+                write!(f, "[")?;
+                for (i, (_, element)) in slots.iter().enumerate() {
+                    let sep = if i == 0 { "" } else { ", " };
+                    write!(f, "{sep}{}", LiteralFmt(element))?;
+                }
+                return write!(f, "]");
+            }
```
ed255, line 250: I think you should add an else here with a warning or an error, or a panic.  Otherwise if an array with gaps arrives here, it won't be rendered, silently.

libs/sdk/src/tests.rs
```diff
@@ -1302,3 +1369,708 @@ fn test_set_guards() {
...
+        }
+"#;
+    let err = match Sdk::default().load_module_from_src_actions(craft_src, &["BadTable"]) {
+        Ok(_) => panic!("expected a var inside a container literal to be rejected"),
+        Err(err) => err.to_string(),
+    };
+    assert!(err.contains("is a var, not a literal"), "{err}");
+}
+
+/// Renders sparse arrays as commitments because Podlang has no sparse-array
```
ed255, line 1520: Add a note about https://github.com/0xPARC/pod2/pull/541

libs/sdk/src/tests.rs
```diff
@@ -1302,3 +1369,708 @@ fn test_set_guards() {
...
+    let _ = env_logger::builder().is_test(true).try_init();
+    let craft_src = r#"
+        fn MintChart(action) {
+            var chart = action.output("Chart");
+            chart.set([["code", 1], ["x", 0]]);
+        }
+
+        fn RevealChart(action) {
+            var chart = action.mutate("Chart");
+            chart.update("x", action.array_get([10, 20, 30], chart.code));
```
ed255, line 1589: Interesting, I guess this is assigned to an anonymous wildcard internally?

libs/sdk/src/lib.rs
```diff
@@ -465,6 +608,47 @@ impl ActionContext {
...
+            return Ok(());
+        }
+        if self.var_state.contains_key(&new) {
+            return Err(format!("var {new} already exists").into());
+        }
+        let state = self.var_state.remove(old).expect("renaming a live var");
+        self.var_state.insert(new.clone(), state);
+        for var in self.vars.iter_mut().filter(|v| v.as_str() == old) {
+            *var = new.clone();
+        }
```
ed255, line 639: we should expect to find `old` only once.  What about making it more clear like this?
```
let old_var = self.vars.iter_mut.find(|v| v.as_str() == old).expect("old exists");
*old_var = new.clone();
```

libs/sdk/src/lib.rs
```diff
@@ -2794,10 +3130,29 @@ fn new_engine() -> Engine {
...
+                    let old_name = match &*arg_ctx.arg.borrow() {
+                        VarOrValue::Value(_) => None,
+                        VarOrValue::Var(var) => Some(var.name.clone()),
+                    };
+                    if let Some(old_name) = old_name {
+                        {
+                            let mut ctx = arg_ctx.ctx.0.borrow_mut();
+                            // Preserve a generated variable's identity and wildcard
+                            // position when the script assigns it a name.
+                            if is_anon_var(&old_name) {
```
ed255, line 3148: my understanding of the code is that if a user defines a `var _foo` then a o reassign it would be detected as anon var and renamed?
There was a comment in `is_anon_var` related to this that I didn't fully understand:
```
/// Script-provided names may also begin with this prefix. Treating them as
/// generated allows a subsequent binding to rename the handle instead of
/// creating an alias.
```

The comment explains this behavior as a feature, but it seems more like a side effect?  Would you say it's advisable for users to **not** prefix a `var` with `_`?

ed255 (approved):
LGTM! I've left a few small comments

libs/sdk/src/tests.rs
ed255: The answer is yes!  Now the sdk creates anonymous wildcards automatically, and they are prefixed with `_`.  On assignment they are renamed to whatever the script says via `var foo = ...`

PR #236, opened 2026-08-31
Avoid building intro pod circuits during SDK module loading

robknight:
Pin the VDF and lt-eq-u256 verifier-data hashes so the SDK can register intro dependencies without initializing their proving circuits. This avoids unnecessary circuit loading during operations such as `pexe inspect plan`.

Add explicit driver warm-up functions and verify each pinned hash when its corresponding circuit is actually loaded or built.


ed255 (approved):
LGTM!

